PR #1, opened 2020-01-27
Add missing binaries when doing a make install

mackoj:


mattt:
Hey @mackoj! Thanks for opening this PR.

The one concern I have with this change is that, when the experimental tools eventually go away (either being subsumed by `swift-doc` or spun out into a separate project), early adopters will end up with outdated binaries polluting their `$PATH`. That is to say, I worry that codifying the install process for something that we don't intend to support long-term sends mixed signals.

What do you think the right balance is here? And out of curiosity: was your PR motivated by your wanting to start using one of these in a workflow?

mackoj:
I do understand your point of view, I did it mainly because I'm working on a new project and wanted the doc to be generated by CI.

Early adopters pay the price of being early adopter 🤷‍♂️.
The semver is **0.0.1** I hope early adopters understand that everything could change at this point.
I think you could let it there and add a **big** warning message.

mattt:
> The semver is **0.0.1** I hope early adopters understand that everything could change at this point.
> I think you could let it there and add a **big** warning message.

This gets into some interesting meta-SemVer territory. Despite it being 0.0.1, what (very likely) isn't going to change is that this package will provide an executable named `swift-doc`. If someone does `make install`, there's a strong expectation that `make uninstall` will remove all traces of the previous action — no matter what version they had installed.

`swift-api-diagram` and `swift-api-inventory` are intended to be temporary and unsupported, and providing a sanctioned installation would seem to run counter to that. If someone really wants to install compiled binaries for these, they could still do so without a `Make` target.

If you still feel strongly about this, what would you say to resolving this with documentation instead?

mackoj:
> If you still feel strongly about this, what would you say to resolving this with documentation instead?

Yes documentation is the best option in this case.

Thanks for your time,

PR #3, opened 2020-01-28
Fix link for markdown files

unnamedd:
# Description
As I user, I want to have the correct filename in the link for another file.

# Problem
The files generated had a correct name but in the generated content in the body of the markdown, the link missed ".md".

Here a little example:

```diff
// from
-[File](File)
// to
+[File](File.md)
```

Based on the changes I made, I would like to know if it make sense for you @mattt.

### DIFF
--- a/Sources/swift-doc/Supporting Types/Components/ConformingTypes.swift
+++ b/Sources/swift-doc/Supporting Types/Components/ConformingTypes.swift
@@ -28,7 +28,7 @@ struct ConformingTypes: Component {
                     #"""
                     \#(names.map { name in
                         if module.hasDeclaration(named: name) {
-                            return "[`\(name)`](\(path(for: name)))"
+                            return "[`\(name)`](\(path(for: name)).md)"
                         } else {
                             return "`\(name)`"
                         }
--- a/Sources/swift-doc/Supporting Types/Components/Inheritance.swift
+++ b/Sources/swift-doc/Supporting Types/Components/Inheritance.swift
@@ -53,7 +53,7 @@ struct Inheritance: Component {
                         #"""
                         \#(inheritance.map {
                             if module.hasDeclaration(named: $0) {
-                                return "[`\($0)`](\(path(for: $0)))"
+                                return "[`\($0)`](\(path(for: $0)).md)"
                             } else {
                                 return "`\($0)`"
                             }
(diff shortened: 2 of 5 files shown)


mattt:
Hi @unnamedd, thanks for submitting this PR. 

From the [README](https://github.com/SwiftDocOrg/swift-doc/blob/master/README.md):

> The CommonMark files generated by swift-doc are formatted for publication to your project's GitHub Wiki

From [GitHub's documentation for editing GitHub Wiki content](https://help.github.com/en/github/building-a-strong-community/editing-wiki-content#adding-links):

> If your pages are rendered with Markdown, the link syntax is `[Link Text](full-URL-of-wiki-page)`

It's not entirely clear from the docs, but the full URL of the wiki page doesn't include the `.md` file extension. With this change, links in the wiki would resolve to their raw source.

I know that CommonMark output is generally useful outside the context of GitHub Wiki, and I intend for `swift-doc` to offer a more platform-agnostic output option in the future. But until it's clearer what that should look like, I'm inclined to keep things as they are.

unnamedd:
Hi @mattt,
I understand your point to follow what is being said at the GitHub’s documentation and also what is written in the README, but the problem I’ve found was related to the mismatching of the file path printed in the document automatically generated and the real file name, swift-doc generate every file with an extension `.md` and the doesn't have the .md and then, the link doesn't work.

I agree that my implementation wasn't really good and I was not really happy with it, I guess we should inject somewhere the file extension used when the file is created, it should prevent this kind of change I did. 

What do you think?

mattt:
I'm sorry, but I'm having some trouble understanding what you mean. To clarify, links in the generated CommonMark files don't include `.md` in their URL strings because of how links in GitHub Wiki pages resolve. I closed this PR because the proposed change would have, by adding `.md` file extensions, caused links in GitHub Wiki pages to resolve to raw files rather than pages, which would break existing behavior.

Is there anything about the current implementation that breaks generation for GitHub wiki pages?

unnamedd:
@mattt,

I'm really sorry for the possible confusion I would bring to you. You are completely right regarding the GitHub Wiki Pages.

When we generate the documentation and push it directly to GitHub Wiki Pages, it works fine. 
The confusion from my side what to think that every other place would do the same thing. When I generated the documentation and opened using Xcode, VisualStudio Code or any other tool, links weren't working because the name of the file contains `.md` and the link doesn't have, but as your current intention is to have it working on GitHub Wiki Pages, then my PR doesn't fit well here.

In any case, thank you for your attention and thanks for the project. 
Hope to have a better contribution soon.

PR #7, opened 2020-01-29
Expose executable and library in swift package

doozMen:
This way you could actually do as instructed in the readme `swift run swift-doc"


mattt:
This has been fixed by 0b49deb.

PR #10, opened 2020-01-29
README: Corrections for three minor typos

morqon:
A "look after the pennies" PR: no major copy edits, just three small typos excised. 

Corrections:

-  "how all everything" –> "how everything"
- "come up with" –> "coming up with"
- "installed both macOS" –> "installed on both macOS"

Thanks, and good luck!


mattt:
Thanks so much for these, @morqon. I ended up rewriting the second correction you pointed out. Let me know if that still isn't reading correctly.

PR #24, opened 2020-03-29
add support for yonaskolb/Mint

thecb4:
Looks to be one line to support https://github.com/yonaskolb/Mint

I'm looking to use this as part of macOS and Linux pipelines. Keeping things as Swift as possible (no gems, homebrew) is a goal.


mattt:
This has been fixed by 0b49debec97c762a86f64de68653fa0aac35d8bd.

Issue #25, opened 2020-04-01
Issues generating HTML documentation for Nuke


kean:
Hi Mattt, this looks absolutely fantastic and I would really like to start using swift-doc in my projects.  I downloaded the latest swift-doc version from `master` (85b5872e3cdb58be273d3dd2315fa101acbfbe3c) and tried running it against [Nuke](https://github.com/kean/Nuke)'s master.

```
swift doc generate ~/Develop/Nuke/Sources/ --output Documentation --format html --module-name Nuke
```

My command line tools are set to Xcode 11.4.

It seems it got a bit confused. I think it might be because of some of the "namespaces" and typealiases.

The actual types are named `ImageProcessors.GaussianBlur`, etc.

<img width="944" alt="Screen Shot 2020-04-01 at 13 13 56" src="https://user-images.githubusercontent.com/1567433/78166526-ee0ca800-741a-11ea-83f5-395ef8eee8f1.png">

[Documentation.zip](https://github.com/SwiftDocOrg/swift-doc/files/4416673/Documentation.zip)

kean:
What I also noticed is that:

- It generated documentation for internal types.
- It got confused with enumerations. For example, for `ImagePipeline/index.html` file it added 12 enumeration cases. They all actually belong to different types.

kean:
[Documentation.zip](https://github.com/SwiftDocOrg/swift-doc/files/4416673/Documentation.zip)

mattt:
Thanks for your feedback, @kean! 

Responding really quick to your first bit, about the incorrect (long) type names: I was seeing the same thing with #16, but thought it was limited to Linux. I'll be really happy if I can reproduce it locally with Nuke; that'll let me troubleshoot the issue _much_ more quickly.

mattt:
@kean, this should be fixed by #51. Can you please try running this again with the latest `master`?

kean:
Thanks, @mattt! Most of the issues have been resolved.

I still observe an issue with nested types.

- There is a nested struct `ImagePipeline.Configuation`. In the generated documentation, all of the fields of this struct are displayed as if they are fields of `ImagePipeline` itself. There are also displayed as fields of `ImagePipeline.Configuation`, but the in-line documentation is gone.

Is it worth opening a separate issue?

kean:
One more note, the HTML links in the documentation don't work if you browse it locally, which is a bit inconvenient.

mattt:
[...]

Issue #31, opened 2020-04-01
Use IndexStoreDB for Symbol Resolution 


mattt:
Currently, swift-doc builds a symbol graph syntactically. However, this approach has limitations (#25, #16). We may want to keep the current, syntactic approach around as an option for quick-and-dirty doc generation, but in the long-term, we'll likely want to leverage [IndexStoreDB](https://github.com/apple/indexstore-db) for this.

mattt:
Having dug into this a bit over the past few days, here's what I've learned:

- As it turns out, our current syntactic approach isn't inherently wrong. It just had a bug in its implementation (#51)
- IndexStoreDB can help, but it's not a panacea. Having USRs for symbols goes a long way to disambiguating relationships — especially across modules, but it's no substitute for the information we're getting from our syntactic parse.

Going forward, I think the best approach will be to take a progressive approach to doc generation. As a foundation, we start with a syntactic read of Swift code. If available, we then layer on information from IndexStoreDB (we might conditionalize certain features like symbol linking on the availability of this information). Finally, we might then pull in information from imported modules, like official docs for the Swift Standard Library or Foundation.

MaxDesiatov:
I wonder if more detailed type information would help in any way. Some time ago I've attempted to implement a [type inference engine](https://github.com/maxdesiatov/typology) for this type of dev tools (documentation generators, refactoring tools) etc, I wonder if it could be of some use. Especially as symbol resolution in Swift is exarcebated by the presence of overloads, shadowed symbols etc

mattt:
Alternatively, we can load [generated symbol graphs](https://github.com/apple/swift/pull/28678).

PR #62, opened 2020-04-08
Fix nested type handling

victor-pavlychko:
First of all, thanks for a great tool! 🙂

While running swift-doc with one of my current projects, I noticed a bug with nested type handling.

Consider the following input:

```swift
public class Outer { }

extension Outer {
    public enum Inner {
        case Case
    }
}

extension Outer.Inner {
    public static let Ver = 0
}
```

This snippet triggers the following issues:
- `Case` and `Ver` symbols are mistakenly associated with `Outer` symbol
- No "nested types" and "member of" sections are generated

### DIFF
--- a/Sources/SwiftDoc/Interface.swift
+++ b/Sources/SwiftDoc/Interface.swift
@@ -55,9 +55,9 @@ public final class Interface: Codable {
     public private(set) lazy var relationships: [Relationship] = {
         var relationships: Set<Relationship> = []
         for symbol in symbols {
-            let owner = symbol.context.last(where: { $0 is Extension || $0 is Symbol })
+            let lastDeclarationScope = symbol.context.last(where: { $0 is Extension || $0 is Symbol })
 
-            if let container = owner as? Symbol {
+            if let container = lastDeclarationScope as? Symbol {
                 let predicate: Relationship.Predicate
 
                 switch container.api {
@@ -74,7 +74,7 @@ public final class Interface: Codable {
                 relationships.insert(Relationship(subject: symbol, predicate: predicate, object: container))
             }
 
-            if let `extension` = owner as? Extension {
+            if let `extension` = lastDeclarationScope as? Extension {
                 if let extended = symbols.first(where: { $0.api is Type &&  $0.id.matches(`extension`.extendedType) }) {
 
                     let predicate: Relationship.Predicate
--- a/Sources/SwiftDoc/Interface.swift
+++ b/Sources/SwiftDoc/Interface.swift
@@ -55,9 +55,9 @@ public final class Interface: Codable {
     public private(set) lazy var relationships: [Relationship] = {
         var relationships: Set<Relationship> = []
         for symbol in symbols {
-            let `extension` = symbol.context.compactMap({ $0 as? Extension }).first
+            let owner = symbol.context.last(where: { $0 is Extension || $0 is Symbol })
 
-            if let container = symbol.context.compactMap({ $0 as? Symbol }).last {
+            if let container = owner as? Symbol {
                 let predicate: Relationship.Predicate
 
                 switch container.api {
@@ -74,7 +74,7 @@ public final class Interface: Codable {
                 relationships.insert(Relationship(subject: symbol, predicate: predicate, object: container))
             }
 
-            if let `extension` = `extension` {
+            if let `extension` = owner as? Extension {
                 if let extended = symbols.first(where: { $0.api is Type &&  $0.id.matches(`extension`.extendedType) }) {
 
                     let predicate: Relationship.Predicate
--- a/Sources/swift-doc/Supporting Types/Components/Relationships.swift
+++ b/Sources/swift-doc/Supporting Types/Components/Relationships.swift
@@ -36,6 +36,23 @@ struct Relationships: Component {
         self.inheritedTypes = module.interface.typesInherited(by: symbol) + module.interface.typesConformed(by: symbol)
     }
 
+    var graphHTML: HypertextLiteral.HTML? {
+        var graph = symbol.graph(in: module)
+        guard !graph.edges.isEmpty else { return nil }
+
+        graph.aspectRatio = 0.125
+        graph.center = true
+        graph.overlap = "compress"
+
+        let algorithm: LayoutAlgorithm = graph.nodes.count > 3 ? .neato : .dot
+
+        do {
+            return try HypertextLiteral.HTML(String(data: graph.render(using: algorithm, to: .svg), encoding: .utf8) ?? "")
+        } catch {
+            logger.error("\(error)")
+            return nil
+        }
+    }
 
     var sections: [(title: String, symbols: [Symbol])] {
         return [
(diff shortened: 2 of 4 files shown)


mattt:
Thanks so much for taking the time to identify this issue and provide a fix, @victor-pavlychko. I really appreciate it!

I'd been holding off on investing in a robust test suite for this functionality, because I thought it might soon be replaced (see #31). But now that I'm more confident that the current, syntactic approach will be sticking around, I'd like to add more unit tests to exercise these kinds of corner cases.

Could you please add a test case to verify the behavior of your patch? The example you used here is great; just a few requested changes to match the conventions of the existing test:

```swift
public class C { }

extension C {
    public enum E {
        case c
    }
}

extension C.E {
    public static let tp = 0
}
```

Sources/swift-doc/Supporting Types/Components/Relationships.swift
```diff
@@ -73,30 +73,36 @@ struct Relationships: Component {
     }
 
     var html: HypertextLiteral.HTML {
-        var graph = symbol.graph(in: module)
-        guard !graph.edges.isEmpty else { return "" }
-
-        graph.aspectRatio = 0.125
-        graph.center = true
-        graph.overlap = "compress"
+        guard sections.map(\.symbols.count).reduce(0, +) != 0 else { return "" }
```
mattt, line 76: I'm still on the fence about Swift's new keypath `map` syntax. But setting that aside: We don't need to compute the total count to determine if there are any symbols. A better way would be to do `let _ = sections.first { !$0.symbols.isEmpty }`.

Sources/swift-doc/Supporting Types/Components/Relationships.swift
```diff
@@ -73,30 +73,36 @@ struct Relationships: Component {
...
 
-        do {
-            svg = try HypertextLiteral.HTML(String(data: graph.render(using: algorithm, to: .svg), encoding: .utf8) ?? "")
-        } catch {
-            logger.error("\(error)")
+            do {
+                svg = try HypertextLiteral.HTML(String(data: graph.render(using: algorithm, to: .svg), encoding: .utf8) ?? "")
+            } catch {
+                logger.error("\(error)")
+            }
```
mattt, line 91: I think this has gotten complex enough to warrant extraction into a separate helper method / property (e.g. `var svg: HyperTextLiteral.HTML`. What do you think?

Sources/swift-doc/Supporting Types/Components/Relationships.swift
```diff
@@ -73,30 +73,36 @@ struct Relationships: Component {
...
         }
 
         return #"""
         <section id="relationships">
             <h2 hidden>Relationships</h2>
-            <figure>
-                \#(svg ?? "")
+                \#(svg.flatMap { svg in
```
mattt, line 97: I'm not sure what `flatMap` is providing here as compared to the original, but that could just be me having trouble reading the diff.

Sources/SwiftDoc/Interface.swift
```diff
@@ -55,9 +55,9 @@ public final class Interface: Codable {
     public private(set) lazy var relationships: [Relationship] = {
         var relationships: Set<Relationship> = []
         for symbol in symbols {
-            let `extension` = symbol.context.compactMap({ $0 as? Extension }).first
+            let owner = symbol.context.last(where: { $0 is Extension || $0 is Symbol })
```
mattt, line 58: I'm not sure `owner` the right name for this. Can we come up with something better? (If not, we can go with something inelegant but precise, like `lastExtensionOrSymbolContext`)

Sources/swift-doc/Supporting Types/Components/Relationships.swift
victor-pavlychko: I'm not sure about checking the first item only:
- I noticed that `sections` are processed with `filter { !$0.symbols.isEmpty }` so there should be no empty sections at the moment, we can just test for `sections.isEmpty`
- at the same time, I see `guard !symbols.isEmpty else { return nil }` in the generation code

So I decided to write a more defensive check to be sure that it won't break in the future.

I'd prefer to state that we never have empty sections and simplify/remove all relevant checks.

Sources/swift-doc/Supporting Types/Components/Relationships.swift
victor-pavlychko: I moved the "<figure>" block inside `flatMap` so it is not rendered when the `svg` data is missing.

Sources/SwiftDoc/Interface.swift
victor-pavlychko: I can also propose `lastDeclarationScope`, what do you think?

Sources/swift-doc/Supporting Types/Components/Relationships.swift
victor-pavlychko: Agree, moved to `graphHTML` property.

victor-pavlychko:
Thanks for the quick feedback. I have added a test for `SwiftDoc` library functionality. As for the `swift-doc` rendering code, the tests are also there, but, I have disabled them for now. It looks like SPM can't test executable targets.

kean:
@victor-pavlychko, I was going to work on the same thing but you beat me to it 😭

> As for the swift-doc rendering code, the tests are also there, but, I have disabled them for now. It looks like SPM can't test executable targets.

I was wondering the same thing. Regardless of whether it can or can't test executables, it might be worth moving all of this functionality to SwiftDoc library.

Sources/SwiftDoc/Interface.swift
mattt: Perfect!

kean:
I performed manual testing, the changes in this branch fixed the issue https://github.com/SwiftDocOrg/swift-doc/issues/25

Sources/swift-doc/Supporting Types/Components/Relationships.swift
mattt: To clarify: the use of `first` doesn't check the first item, but returns as soon as it finds a section with non-empty `symbols`. But taking another look, you're right: a simple check on `sections.isEmpty` should be sufficient. The other guard statement you mentioned was from an earlier iteration, and I neglected to remove it.

Sources/swift-doc/Supporting Types/Components/Relationships.swift
mattt: Ah, got it. Thanks for clarifying.

mattt:
> @kean: Regardless of whether it can or can't test executables, it might be worth moving all of this functionality to SwiftDoc library.

I'm still not 100% on the current shape and spelling of the SwiftDoc module, but I'm pretty confident that anything to do with rendering pages should be in a separate module.

While we're waiting on end-to-end tests for the executable, could we write tests that `@testable import swift-doc`?  (I don't think I've ever tried to do anything with the module of an executable, so I'm genuinely curious). If not, then I think the next best thing for now would be to move everything but the commands themselves into a new module, and test the rendering behavior through that directly.

victor-pavlychko:
> @mattt: While we're waiting on end-to-end tests for the executable, could we write tests that `@testable import swift-doc`? (I don't think I've ever tried to do anything with the module of an executable, so I'm genuinely curious).

Tried that: it complies swift files and then fails to link with the `swift-doc` module throwing a bunch of unresolved external errors.

I suggest leaving those test methods disabled for now and enabling them once the functionality is moved somewhere.

Sources/swift-doc/Supporting Types/Components/Relationships.swift
victor-pavlychko: Oh, right, I've misunderstood your initial suggestion. Thanks for clarifying that.

Changing to `isEmpty` + removed that `guard` statement.

Tests/SwiftDocTests/NestedTypesTests.swift
```diff
@@ -0,0 +1,127 @@
...
+        XCTAssert(`enum`.api is Enumeration)
+
+        let classRelationships = Relationships(of: `class`, in: module)
+        XCTAssertNotEqual(classRelationships.html, "")
+
+        let enumRelationships = Relationships(of: `enum`, in: module)
+        XCTAssertNotEqual(enumRelationships.html, "")
+    }
+
+    func testNoRelationshipsSeciont() throws {
```
mattt, line 98:
```suggestion
    func testNoRelationshipsSection() throws {
```

Sources/swift-doc/Supporting Types/Components/Relationships.swift
```diff
@@ -36,6 +36,23 @@ struct Relationships: Component {
         self.inheritedTypes = module.interface.typesInherited(by: symbol) + module.interface.typesConformed(by: symbol)
     }
 
+    var graphHTML: HypertextLiteral.HTML? {
```
mattt, line 39: Thanks for extracting this into a property. FYI: I'll probably end up further extracting this (along with the enclosing `<figure>` tag) into a new component as part of #67.

Tests/SwiftDocTests/NestedTypesTests.swift
victor-pavlychko: Thanks 🤦‍♂️

mattt:
Perfect. Thanks again, @victor-pavlychko! 🎉

Issue #56, opened 2020-04-05
Path-relative URLs are not working when deployed to GitHub Pages


kean:
Hi, I [deployed](https://kean-org.github.io/docs/nuke-docs-beta/9.0.0) the documentation generated by swift-doc to GitHub pages. There is an issue with path-relative URLs. For example, if you got to [ImageProcessing](https://kean-org.github.io/docs/nuke-docs-beta/9.0.0/ImageProcessing/) page and click of any of the URLs there, e.g. `ImageProcessors.Resize`, it shows 404 page. 

I'm using the latest swift-doc from `master`:

```
commit 0b2291a52f22a30d5e2528f06be79d40cac570cb (HEAD -> master, origin/master, origin/HEAD)
Merge: e5c4b85 0ac5744
Author: Mattt <mattt@me.com>
Date:   Sun Apr 5 08:15:44 2020 -0700

    Merge pull request #53 from SwiftDocOrg/mattt/operator-fix
    
    Fix logic error when classifying operators in generate subcommand
```

mattt:
Hey @kean. This is a known, albeit undocumented limitation, so thanks for reporting it.  Currently, all links are routed using absolute paths, which works for sites hosted at a root domain, but not when they're located in a subdirectory, as in your case.

To solve this correctly, we'll likely need to add a new (optional) parameter to the `generate` subcommand to specify a base URL. If one is specified, it could either be passed to a [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) tag in the final HTML, or factored into the `path` helper method.

kean:
Have you considered using using Relative URLs? Something like this should work:

```
<a href="../ImageProcessors_Resize">ImageProcessors.Resize</a>
```

mattt:
As currently implemented, relative links would work on symbol pages, but not for the Home page. 
For me, relative paths rank below the two other approaches I described, primarily because it's dependent on a fixed page routing behavior, which we may decide to change in the future.

PR #65, opened 2020-04-09
Add baseUrl option

kean:
Fixes https://github.com/SwiftDocOrg/swift-doc/issues/56

- Add a new `--base-url` (`baseURL: String`) option
- Update HTML generation to use the new `baseUrl` option
- Update generated links to be relative

Usage:

```
swift doc generate ~/Develop/Nuke/Sources/ \
    --output Documentation \
    --format html \
    --module-name Nuke \
    --base-url https://kean-org.github.io/docs/nuke-docs-beta/9.0.0/
```

## Notes

I made the minimum number of changes needed to make https://kean-org.github.io/docs/nuke-docs-beta/9.0.0/ work. If you don't think this is the right direction, feel free to close it. I'll keep my workaround locally for now.

What I'm not happy about in this solution is that it is not going to work if you browse it locally. Jazzy uses relative URLs in a form of `<a href="../ImageCaching">`. There are two major advantages. It works regardless of how you deploy it (root or not). And it works locally.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `--base-url` option.
+  #65 by @kean.
+
 ## [1.0.0-beta.2] - 2020-04-08
 
 ### Changed
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -29,6 +29,11 @@ extension SwiftDoc {
                     default: .commonmark,
                     help: "The output format")
             var format: Format
+
+            @Option(name: .shortAndLong,
+                    default: "/",
+                    help: "The base URL where the HTML documentation is going to be hosted")
+            var baseUrl: String
         }
 
         static var configuration = CommandConfiguration(abstract: "Generates Swift documentation")
@@ -79,7 +84,7 @@ extension SwiftDoc {
                     }
 
                     let url = outputDirectoryURL.appendingPathComponent(filename)
-                    try page.write(to: url, format: format)
+                    try page.write(to: url, format: format, baseUrl: options.baseUrl)
                 } else {
                     switch format {
                     case .commonmark:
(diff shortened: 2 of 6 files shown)


mattt:
Great work, @kean! Thanks so much for taking this on.

I pushed a few (small) changes to your PR branch, all to do with naming and documentation. Please take a look at those and let me know if you have any objections to them.

> What I'm not happy about in this solution is that it is not going to work if you browse it locally. Jazzy uses relative URLs in a form of `<a href="../ImageCaching">`.

Browsing locally — that is opening `file:///` URLs in Safari — has a lot of caveats. Safari imposes various restrictions on what resources can be loaded / what scripts can be executed for local HTML files (see "Develop > Disable Local File Restrictions"), while macOS itself has ratcheted up security in recent years, requiring permissions for apps to get read access to directories. And that's just one browser / OS combination. macOS / iOS / Linux / Windows. × Safari / Chrome / Firefox / Edge... the combinatorics are overwhelming.

As we increase the capabilities of the HTML output, it'll be difficult enough to guarantee common behavior for normal web standards stuff. I'd hate to layer on the additional complexity of handling `file:///` URLs when you can easily start a local web server.

kean:
Thanks, @mattt, I agree with your changes. Updated changelog.

> As we increase the capabilities of the HTML output, it'll be difficult enough to guarantee common behavior for normal web standards stuff.

You make a good point, I agree with you. It's not going to be feasibly to keep everything working locally when more features are added. And it's not worthwhile, nobody browses these files locally, probably except me.

Changelog.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## TBD
```
mattt, line 10:
```suggestion
## [Unreleased]
```

Changelog.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Added
+
+- Added `--base-url` option.
+ #65 by @kean.
```
mattt, line 15:
```suggestion
  #65 by @kean.
```

mattt:
Thanks for adding that Changelog entry. Just a couple formatting fixes (I was hoping to commit from the suggestions, but that wasn't working for the header deletion)

kean:
No worries, I push the updated changelog. Also squashed the formatting changes into the same commit.

mattt:
Alright, merged! Thanks again for your help, @kean 🎉

Issue #67, opened 2020-04-09
Rearchitect documentation generators


mattt:
One of the consequences of #65 is that we now have an option that applies to HTML output, but not CommonMark. This kind of divergence isn't unexpected — certainly, as we expand functionality to support additional output formats like DocSets (#27), it was inevitable.

We're at the point now where the original assumptions we made to ship the initial proof-of-concept are starting to break down, and it'd be great to find a solution that will set us up for success in version 1.x and beyond.

Here's what I'm looking for in a solution, in order of importance:

1. Something that lets us add support for new formats, without encroaching on or breaking the existing ones.
2. The ability to test generated output without actually running the `swift-doc` command (a full integration test should be run separately), as discussed in [#62](https://github.com/SwiftDocOrg/swift-doc/pull/62#issuecomment-611200909)
3. The ability for users to use a 3rd-party generator without forking `swift-doc` itself — for extensibility, but also some flexibility for us about what's included in the main offering.

...so basically, a plugin system. 😭

So far, the only solution I've been able to come up with that meets all three of these criteria is to make each format its own subcommand of `generate`. For example, running:

```swift
$ swift doc generate html
```

...would look for an executable named `swift-doc-generate-html`. This is the same approach that allows `swift-doc` to be executed as a subcommand of `swift`, and would let anyone write and/or install their own custom output format.

Does anyone else have any other ideas about what we should do?

muizidn:
Would love to see and help implement this feature even if first iteration isn't optimal yet. My use case for now is I need a PDF generator which allow me to send fully cross referenced PDF document to the interested client.

Anything I can start from @mattt?

mattt:
@muizidn I wouldn't consider PDF to be a primary output format for `swift-doc`. Instead, I'd recommend using a tool like [PrinceXML](https://www.princexml.com) to generate PDF from HTML + CSS.

muizidn:
Ah sure. Thanks your reply.

PR #76, opened 2020-04-10
Fixing protocol requirements html rendering

### LINKED ISSUE #71: Protocol members are not displayed in the generated HTML
Requirements do not appear in the HTML output but are present in the markdown version. Please let me know if you want me to open a PR for the fix.

victor-pavlychko:
Fixes #71 

This PR aligns `Requirements` with other similar components: `Members` and `Relationships`. What do you think about extracting shared section logic into a new component or base class?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   #62 by @victor-pavlychko.
 - Fixed rendering of type relationships section when no graph data is available.
   #62 by @victor-pavlychko.
+- Fixed rendering of protocol requirements in the HTML version.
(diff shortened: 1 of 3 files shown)


mattt:
Great work, @victor-pavlychko! Thanks for taking this on.

> What do you think about extracting shared section logic into a new component or base class?

Yeah, this definitely seems like a useful pattern that we may want to apply more systematically. I'll factor this into my approach for #67.

victor-pavlychko:
Sure, I added one more entry.

I have also checked my previous PR with GraphViz installed and noticed a bug with HTML figure escaping, adding the fix here too, as it's a one-liner.

mattt:
Awesome. Thanks again, @victor-pavlychko! Merging this now.

PR #79, opened 2020-04-11
Fix crash #78

### LINKED ISSUE #78: malloc crash “randomly”
Hey!

When trying to generate docs for the libraries in an app I’m working on I keep running into `malloc` crashes. This happens for different libraries for different runs. Both when using the markdown and HTML formats.

Basically I run this command for a number of modules:

```shell
swift doc generate \
  --output Libraries/$module_name \
  --module-name $module_name \
  --output artifacts/docs/API/$module_name \
  [--format html]
```

The errors I see are:

```
swift-doc(44797,0x70000879f000) malloc: Double free of object 0x7ff8b6d11c00
swift-doc(44797,0x70000879f000) malloc: *** set a breakpoint in malloc_error_break to debug
```

```
swift-doc(50645,0x700003d39000) malloc: *** error for object 0x7ff087124ba0: pointer being freed was not allocated
swift-doc(50645,0x700003d39000) malloc: *** set a breakpoint in malloc_error_break to debug
```

For now I’ve worked around it by retrying the `swift doc` operation on failures. Which seems to generate all docs.

---

Side note, I tried to build the project in Xcode to help debug the issue. Unfortunately Xcode (11.4 (11E146)) fails to build with an error:

```
Swift Compiler Error Group
/Applications/Xcode-1140.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/libxml2/libxml/module.modulemap:1:8: Redefinition of module 'libxml2'
/Users/.../swift-doc/.build/checkouts/Markup/Modules/module.modulemap:1:8: Previously defined here
```

I can file a separate issue for this if you want?

kean:
Fixes #78.

The root cause is that `Interface` is not thread safe because of all of the lazy properties.

I measured the performance before and after on my project with 68 pages.

Before: **90 ms**.
After: **100 ms**.

This operation could probably be optimized by writing from 2-3 threads, but not from 68 `DispatchQueue.concurrentPerform` iterations. I'm not sure it is worth optimizing it in the first place.

### DIFF
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -95,19 +95,19 @@ extension SwiftDoc {
                         pages["Home"] = HomePage(module: module)
                     }
 
-                    try pages.map { $0 }.parallelForEach {
+                    for (name, page) in pages {
                         let filename: String
                         switch format {
                         case .commonmark:
-                            filename = "\($0.key).md"
-                        case .html where $0.key == "Home":
+                            filename = "\(name).md"
+                        case .html where name == "Home":
                             filename = "index.html"
                         case .html:
-                            filename = "\($0.key)/index.html"
+                            filename = "\(name)/index.html"
                         }
 
                         let url = outputDirectoryURL.appendingPathComponent(filename)
-                        try $0.value.write(to: url, format: format, baseURL: options.baseURL)
+                        try page.write(to: url, format: format, baseURL: options.baseURL)
                     }
                 }
             } catch {


mattt:
Thanks for looking into this, @kean. I'm glad that we seem to have a handle on what was causing this race condition (my best guess was that it had something to do with SwiftSyntax, since it had been causing similar issues in its Swift 5.1 version).

A while back, when I was benchmarking `swift-doc` on a larger project (~500k loc / 10k files / 20k symbols), parallelizing this step dropped the runtime from ~10 minutes to <1 minute. For <100 pages, you're just about breaking even with the coordination cost of `concurrentPerform`.

Long term, it'd be great for us to add some performance tests to our integration process. In the short term, I'd like to see if we can resolve #78 by making `Interface` thread-safe. This may be as simple as converting the `lazy` properties to stored properties initialized with self-executing closures.

PR #80, opened 2020-04-11
Improve parallelMap

kean:
- Remove `withoutActuallyEscaping`, `concurrentPerform` already accepts a nonescaping closure
- Avoid sorting results as it might become quite expensive with large N


mattt:
Thanks for this patch, @kean!

You're totally right about `withoutActuallyEscaping` being unnecessary here. I was less confident about the correctness of preallocating a fixed-size collection to collect the transformed results (hence my original approach), but after re-reading [the Swift Type Layout documentation](https://github.com/apple/swift/blob/master/docs/ABI/TypeLayout.rst) and checking a few assumptions with `MemoryLayout` in a Playground, I can't think of any cases where this wouldn't work.

I don't think we need a Changelog entry for this PR since this is an implementation detail, so I'll merge this now.

PR #81, opened 2020-04-11
Display globals on the home page

### LINKED ISSUE #69: Globals are not displayed in the generated HTML
Operators, functions, typealiases appear in Markdown, but not in HTML.

kean:
Fixes #69.

You can see end-result here: https://kean-org.github.io/docs/nuke-docs-beta/9.0.0/


mattt:
Thanks for fixing this, @kean! Really glad to be seeing globals in the HTML output now.

Issue #72, opened 2020-04-10
Add option to generate subcommand for specifying which access levels are included 


mattt:
My original motivation for `swift-doc` was to generate documentation for libraries. However, as the tool gains wider adoption, teams will likely want to use it internally for their own apps. The major difference between libraries and apps is that apps are typically much less concerned about ACL (if all of your code is in a single app target, everything _"just works"_).

Rather than force teams to change their code in a way that would otherwise provide no immediate benefit (i.e. sprinkling `public` throughout their code base), I think it would make sense to provide an option for `internal` declarations to be included in the generated output.

As far as how this should work: 

- The current behavior of only looking for `public` or `open` declarations should probably continue to be the default. 
- To help folks who might be confused about nothing happening when they run `swift doc generate` (#70), we should communicate something like "No public declarations found, set the `(option)` flag to...). 
- To accommodate future distinctions, this option should probably take an enumerated value rather than a Boolean. So `--minimum-access-level internal` instead of `--include-internal-declarations`.
- And, of course, this should be documented. I think a "Getting Started" guide should be considered a P1 for our 1.0.0 release. (#73)

byron22seven:
Hi 👋 
Has there been any movement on this? 😃 

I see there's an open PR with some conflicts... Would be really helpful 😊

PR #91, opened 2020-04-20
Option for specifying which access levels are included

johankool:
Implements #72 

Not yet tested very extensively.


Sources/SwiftDoc/Symbol.swift
```diff
@@ -58,7 +58,26 @@ public final class Symbol {
 
         return false
     }
-
+    
+    public var isInternal: Bool {
```
Lukas-Stuehrk, line 62: I think either the implementation or the property name is highly misleading. With this implementation, a public property would return `true`. Same for an `open` or `fileprivate` property.

Also, the actual decision whether a symbol is internal is way more complicated. There are quite some cases to consider, especially with extensions. You can take a look at the existing implementation of `isPublic` where these cases are considered.

Sources/SwiftDoc/Interface.swift
```diff
@@ -5,10 +5,12 @@ import struct SwiftSemantics.Protocol
 public final class Interface: Codable {
     public let imports: [Import]
     public let symbols: [Symbol]
+    public let minimumAccessLevel: AccessLevel
```
Lukas-Stuehrk, line 8: In my point of view, adding this property on the interface mixes concerns. The setting for _creating the documentation_ should not be on a structure that describes source code. 

I'd suggest we don't include it as a property, but instead only pass the symbols that have the correct ACL. Same goes for the extension of `Report`. Alternatively we could collect and work with all symbols and then filter when we generate the documentation. But I did not fully think this through, so there might be some implications I did not consider.

But I also think @mattt is the person to decide on this.

Sources/SwiftDoc/API.swift
```diff
@@ -55,3 +55,9 @@ extension Subscript: API {
 
 extension Typealias: API {}
 extension Variable: API {}
+
+public enum AccessLevel: String, Codable {
```
Lukas-Stuehrk, line 59: One fancy thought: Instead of using `String` as the raw representable, we could also use `Int` and make the enum conform to `Comparable`. Then we could do nice expressions like `if minimumAccessLevel >= accessLevel`.

MaxDesiatov: BTW, if it goes that route and supports `Comparable`, should have a note added that it won't be needed with Swift 5.3 thanks to [SE-0266](https://github.com/apple/swift-evolution/blob/master/proposals/0266-synthesized-comparable-for-enumerations.md).

Sources/SwiftDoc/API.swift
```diff
@@ -55,3 +55,9 @@ extension Subscript: API {
...
+    case `public`
+    case `internal`
+    case `private`
```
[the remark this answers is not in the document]
mattt: Personally, I see `fileprivate` as being equivalent to `private` for the purposes of `swift-doc`, in the same way that `public` implies `open` (though admittedly, I don't fully understand the use case for documenting `private` declarations, so maybe I'm missing a key distinction...)

mattt:
Thanks for taking this on, @johankool. I'm taking a look at this now.

Sources/SwiftDoc/API.swift
mattt: I've already started work with the latest 5.3 snapshot to get [package resources](https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md) working, so I'll take a look at using automatic `Comparable` synthesis with an explicit implementation for `<` in a conditional compilation block.

Sources/SwiftDoc/Interface.swift
mattt: I agree with @Lukas-Stuehrk here. An `Interface` should represent a fixed collection of symbols and relationships. For instance, if we were to generate the public API for a library on macOS vs. Linux, we'd associate a separate `Interface` value for each platform.

As to where that determination should be made, that remains an open question. Like Lukas said, it could be done either in "model" layer (by adding a `minimumAccessLevel` parameter to `init(sourceFiles:)`) or in the "view" layer (by having each page / component filter symbols according to a `minimumAccessLevel` property). The former would be simpler, but I'm not 100% sure whether we can correctly reconcile relationships without knowledge of every symbol. For example, I know that a public class can't inherit from a private class, but I haven't thought through all of the possible interactions between protocols, types, and members.

mattt: (@johankool Don't worry about trying to make sense of my rambling commentary here. I'll provide more actionable feedback in my review)

Sources/SwiftDoc/Symbol.swift
mattt: .@Lukas-Stuehrk is correct. Determining access levels for symbols is extremely complicated in Swift. I've had to fix the implementation for `isPublic` a few times already to accommodate unforeseen edge cases (e.g. cases of public enumerations are always public, even though they never have access modifiers).

Sources/SwiftDoc/Interface.swift
johankool: Agreed. It didn’t feel quite right to put it here. Not quite sure what the right place would be though.

Sources/SwiftDoc/API.swift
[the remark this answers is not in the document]
johankool: I guess making distinction between `private` and. `fileprivate` is indeed unneeded for this. It think normally I’d want the internal methods of my app documented and wouldn’t care so much about the private methods. But perhaps some apps have sufficient complexity that also documenting what private methods do is useful.

bielikb:
Hi, this looks interesting, is there any detail/requirement missing to move forward with this work?

muizidn:
Hello! any update on this feature?

johankool:
@muizidn Sorry, no, I haven't looked at this anymore.

johankool:
I think this is now properly fixed in #219.

Issue #86, opened 2020-04-18
example in README for publishing to wiki is failing


heckj:
I wanted to try out the Github action that publishing the docs in commonmark to the github wiki - I've added the job to my pipeline, and it runs "successfully", but when you look at the detail it's failed to do it's work.

The job is available at https://github.com/heckj/OpenTelemetryModels/pull/7/checks?check_run_id=598253694

with the error detail:
```
  with:
    path: Documentation
  env:
    GH_PERSONAL_ACCESS_TOKEN: ***
/usr/bin/docker run --name c1a2251720909414cf48e808fac8244e4c4_b9fe73 --label 430c1a --workdir /github/workspace --rm -e GH_PERSONAL_ACCESS_TOKEN -e INPUT_PATH -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/OpenTelemetryModels/OpenTelemetryModels":"/github/workspace" 430c1a:2251720909414cf48e808fac8244e4c4  "Documentation"
Initialized empty Git repository in /tmp/ci-XXXXkFhEhb/.git/
fatal: remote error: access denied or repository not exported: /3/nw/38/4c/f5/266259832/255487748.wiki.git
On branch master

Initial commit

nothing to commit
error: src refspec master does not match any
error: failed to push some refs to 'https://***@github.com/heckj/OpenTelemetryModels.wiki.git'
```

[...]

PR #92, opened 2020-04-22
super minor tweak to README & warning emit in generate

heckj:
this may be me just being an idiot, but I completely missed `source` vs. `sources` in the README example for using the wiki code generation example, and burned hours hunting it down. For the example, I think it would be awesome if it matched up with the default so that others don't stub their toes on this particular nit.

I also made a minor tweak (separate commit) to emit a warning so that at least some output appears when it fails to find any source files to generate against.

related to #86

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   #62 by @victor-pavlychko.
 - Fixed rendering of protocol requirements in the HTML version.
   #76 by @victor-pavlychko.
-- Fixed defaulty location of sources reference in README
+- Fixed default location of sources reference in README
   #92 by @heckj
 
 ## [1.0.0-beta.2] - 2020-04-08
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed README to clarify use of `swift-doc` vs. `swift doc`
   on the command line.
   #89 by @mattt.
+- Changed the generate command to emit a warning if no source
+  files are found.
+  #92 by @heckj
 
 ### Fixed
 
@@ -30,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   #62 by @victor-pavlychko.
 - Fixed rendering of protocol requirements in the HTML version.
   #76 by @victor-pavlychko.
+- Fixed defaulty location of sources reference in README
+  #92 by @heckj
 
 ## [1.0.0-beta.2] - 2020-04-08
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed README to clarify use of `swift-doc` vs. `swift doc`
   on the command line.
   #89 by @mattt.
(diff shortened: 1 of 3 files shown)


mattt:
@heckj I'm so glad to hear that you finally figured that out, though I'm obviously not thrilled by how much time you had to spend chasing that down. I think the proposed changes to the documentation and code is absolutely spot-on, and will go a long way to helping others avoid having the same issue.

README.md
```diff
@@ -238,7 +238,7 @@ jobs:
       - name: Generate Documentation
         uses: SwiftDocOrg/swift-doc@master
         with:
-          inputs: "Source"
```
mattt, line 241: Ugh. How long has this been here? (That's purely a hypothetical question; I'm too lazy to `git blame` this right now)

I take full responsibility for this. Though if you're looking for a more impersonal agent to rage against, please direct your ire to Alamofire, which was used as my first example and remains the only library I've seen using `Source` instead of `Sources`. Actually, wait... was _I_ the one who decided that? (Again, hypothetical question)

Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -72,7 +72,10 @@ extension SwiftDoc {
           pages[path(for: name)] = GlobalPage(module: module, name: name, symbols: symbols)
         }
 
-        guard !pages.isEmpty else { return }
+        guard !pages.isEmpty else {
+            logger.warning("No pages were found to render.")
```
mattt, line 76:
```suggestion
            logger.warning("No public API symbols were found at the specified path. No output was written.")
```

mattt: Language to be updated when we support variable ACL (#91). Feel free to pitch any alternatives you can think of. I just wanted to be a bit more explicit about cause and effect in the error message.

mattt:
Oh, one more thing – please add lines to the Changelog under the section "Fixed" for the README change and under "Updated" for the logger message change.

Sources/swift-doc/Subcommands/Generate.swift
heckj: WFM!

Changelog.md
```diff
@@ -30,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   #62 by @victor-pavlychko.
 - Fixed rendering of protocol requirements in the HTML version.
   #76 by @victor-pavlychko.
+- Fixed defaulty location of sources reference in README
```
MaxDesiatov, line 36:
```suggestion
- Fixed default location of sources reference in README
```

Changelog.md
```diff
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed README to clarify use of `swift-doc` vs. `swift doc`
   on the command line.
   #89 by @mattt.
+- Changed the generate command to emit a warning if no source
```
MaxDesiatov, line 24:
```suggestion
- Changed the `generate` command to emit a warning if no source
```

mattt:
@heckj Perfect. Thanks again for your help. This goes a long way to making `swift-doc` more usable. I'll merge this in as soon as CI gives me a ✅ .

@MaxDesiatov Gosh, how I love GitHub's "suggestions in PRs" feature. Thanks for jumping in on these!

heckj:
Thanks for the updates last night! I crashed hard.

Issue #82, opened 2020-04-13
Sidebar links are broken in the HTML version


victor-pavlychko:
All sidebar navigational links point to the root document. Looks like this is caused by the `<base href="\#(baseURL)"/>` element.

mattt:
Yeah, this is something @kean pointed out in a recent thread (still catching up from the weekend, sorry!) 

This behavior of base tags surprised me – I didn’t know about this behavior before. In light of this new information, I think we should consider the alternative approach discussed in #65, which was to factor the base URL into the path helper.

kean:
My bad, I misunderstood how <base> works. It applies not just to relative URLs, but also anchors https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base. I use absolute paths on my Jekyll site. It is convenient when this parameter available globally so every page you generate has access to it.

mattt:
@kean Same here wrt/Jekyll and absolute paths. My suggestion to use a `<base>` tag was out of left field, and a bad call on my part. The good news is that we haven't released a new version with that yet, so we can make this change without worrying about backwards compatibility.

dsmurfin:
I think this is also the same issue in the CommonMark version.

![Screenshot 2020-04-22 at 12 29 52](https://user-images.githubusercontent.com/4533485/79976772-12e1c180-8495-11ea-901b-fdb32303bb7b.png)

mattt:
@dsmurfin That is indeed the cause. I'll prioritize a fix for this work to get that out later today.

mattt:
This is fixed by #93.

PR #93, opened 2020-04-24
Reimplement base URL support without <base> tag

mattt:
This PR seeks to reimplement base URL support added in #65 based on feedback from #82. With my sincere apologies to @kean for leading him down the wrong path by advocating for `<base>` without being aware of its limitations.

I fully acknowledge the jankiness of this approach. Tacking on parameters all over the place like this is unsustainable, but I want to get a fix out sooner rather than later. Long-term, this is the kind of configuration I'd like to encapsulate by a `Generator` (#67)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `--base-url` option.
-  #65 by @kean.
+  #65 by @kean and #93 by @mattt.
 - Added asset pipeline for CSS assets.
   #49 by @kaishin.
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ $ make install
 
     OVERVIEW: Generates Swift documentation
 
-    USAGE: swift-doc generate [<inputs> ...] --module-name <module-name> [--output <output>] [--format <format>]
+    USAGE: swift-doc generate [<inputs> ...] --module-name <module-name> [--output <output>] [--format <format>] [--base-url <base-url>]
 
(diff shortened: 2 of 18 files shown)


kean:
Hey, @mattt. Thanks for addressing this. I was going to work on it, but I've been sick for the past two weeks.

kean:
@mattt , CSS also needs to use base-url. Currently it is not working when deployed in non-root folder.

```
<link rel="stylesheet" type="text/css" href="/all.css" media="all" />
```

kean:
Hmm, it seems swift-doc also used --base-url as an output path which was unexpected and I had to fix manually.

Screen Shot 2020-04-27 at 18 02 58

kean:
Ah, I see, the expectation is to put a single all.css in my root directory? I could do that, but I'm not sure how well it's going to work with my personal workflow. I deploy documentation once per version and then never touch it again. This isn't going to work if all.css file keeps changing with potentially breaking changes to the previously generated documentation.

mattt:
> Hmm, it seems swift-doc also used --base-url as an output path which was unexpected and I had to fix manually.

That's definitely a bug. Thanks for pointing that out.

> Ah, I see, the expectation is to put a single all.css in my root directory? I could do that, but I'm not sure how well it's going to work with my personal workflow. I deploy documentation once per version and then never touch it again. This isn't going to work if all.css file keeps changing with potentially breaking changes to the previously generated documentation.

The idea is for `swift doc generate` to create a "bundle", such that everything within that bundle is  sealed and self-sufficient. There may be changes in the DOM or CSS between releases that change appearance, but any generated output will be inert and unaffected by any such changes.

Issue #95, opened 2020-04-28
CSS doesn't link when generating default HTML format


captainbarbosa:
Hi @mattt! Great project. I was poking around and testing the `--format html` option and noticed the `all.css` file isn't linking properly to the generated `index.html` file.

I tried debugging this myself, first by checking `index.html` where it _seems_ to be linked correctly (however I'm not a web developer):

```html
<link rel="stylesheet" type="text/css" href="all.css" media="all" />
```

But when I check out the network console in my browser, it looks like `all.css` is failing:

![image](https://user-images.githubusercontent.com/10850812/80446266-3fe60800-88cb-11ea-947e-5b75ca830a2c.png)

Any idea what could be happening? I'm using 1.0.0-beta.2.

captainbarbosa:
Oh, I think I may have found the cause.

It appears this `base-href` line is causing the file search path for `all.css` be incorrect:

`Index.html`
```html
<head>
    ...
    <base href="/"/>
    <link rel="stylesheet" type="text/css" href="all.css" media="all" />
</head>
```

Changing this from `<base href="/"/>` to `<base href=""/>` fixes the issue, so I think we just need to swap the ordering of [these two lines](https://github.com/SwiftDocOrg/swift-doc/blob/master/Sources/swift-doc/Supporting%20Types/Layout.swift#L14-L15):

```html
<base href=""/>
<link rel="stylesheet" type="text/css" href="all.css" media="all" />
```

I created a pull request for this in https://github.com/SwiftDocOrg/swift-doc/pull/96.

captainbarbosa:
https://github.com/SwiftDocOrg/swift-doc/pull/93 is likely a better solution than this PR, so I'll close this in favor 👍

mattt:
Thanks again for bringing this to our attention, @captainbarbosa. I'll keep this open until after #93 is merged, so that you can confirm the fix for yourself.

BilalReffas:
Hi @mattt thanks for this great project! 
Just wanted to raise that I'm running into this issue as well. I have ´swift-doc 1.0.0-beta.3´ installed on my machine.

PR #96, opened 2020-04-28
Fix link to CSS in default HTML output

captainbarbosa:
This ensures `all.css` is visible by the HTML file before the base URL is set, and fixes https://github.com/SwiftDocOrg/swift-doc/issues/95.

This was the most minimal change I could think of to get this to work, but another option could be to change the default `base-url` argument from `"/"` to just an empty string `""` - let me know if that is preferred and I can make that change.

### DIFF
--- a/Sources/swift-doc/Supporting Types/Layout.swift
+++ b/Sources/swift-doc/Supporting Types/Layout.swift
@@ -11,8 +11,8 @@ func layout(_ page: Page, baseURL: String) -> HTML {
         <meta charset="UTF-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <title>\#(page.module.name) - \#(page.title)</title>
-        <base href="\#(baseURL)"/>
         <link rel="stylesheet" type="text/css" href="all.css" media="all" />
+        <base href="\#(baseURL)"/>
     </head>
     <body>
         <header>


mattt:
Thanks for looking into this, @captainbarbosa! We're actually in the process of switching away from `<base>` (#93), so that's probably going to be the long-term fix. Since you have fresh eyes on the situation, do you see any compelling reasons to stick with `<base>` or have any thoughts about the best alternative approach?

captainbarbosa:
Oh, I didn't notice that. #93 looks like a better solution. I can close my PR in favor of a more long-term solution.

I'm not a regular web developer, but one additional thing I noticed is by setting `<base>`, sub-pages don't reference `all.css` relative to the rest of the project. So for example, the `Docs/Index.html` can find the CSS file, but `Docs/MyClass/Index.html` tries to reference `all.css` and not `../all.css`. This issue occurs after applying #95. I think this is already captured in https://github.com/SwiftDocOrg/swift-doc/pull/93#issuecomment-620257837 so I won't open an issue.

I looked at Jazzy, a similar project, to see if they used `<base>` tags - it doesn't look like they do. I believe the equivalent config setting for that project is [`--root-url`](https://github.com/realm/jazzy/blob/2a1c339ce74737c34e3ea8d80cdeb71c9c24bb3e/lib/jazzy/config.rb#L310-L313). It appears they are [appending as many `../` that are needed](https://github.com/realm/jazzy/blob/5fcab192f34035f4611b11a81c5b926a53a299ac/lib/jazzy/doc_builder.rb#L94) to build out the links for each page, so this might be something to keep in mind when addressing #93.

captainbarbosa:
https://github.com/SwiftDocOrg/swift-doc/pull/93 is likely a better solution than this PR, so I'll close this in favor 👍

PR #97, opened 2020-05-01
Include symbol summaries on Home page in CommonMark format

mattt:
The current version of the Home pages links to each top-level symbol in a bulleted list, with only the symbol's (qualified) name provided. This presentation isn't very information-dense or informative at all.

This PR offers an alternative take that includes quick summaries alongside each symbol. Here's a look of what that looks like for Alamofire:

- [Before](https://github.com/SwiftDocOrg/Alamofire/wiki/Home/3b49cc44e104707d6f9c432cc60c4e1aaa2254a3)
- [After](https://github.com/SwiftDocOrg/Alamofire/wiki/Home/edb99034ceeb198818c5c64ffa639c15c253f295)

This implementation replaces the bulleted list with a definition list. Because those aren't supported in CommonMark (or [GitHub Flavored Markdown](https://github.github.com/gfm/)), it uses raw HTML (`<dl>` / `<dt>` / `<dd>`) tags. However, this isn't a necessary change; if we decide to keep CommonMark output format relatively free of HTML, we could certainly reimplement with a bulleted list instead.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,9 +5,9 @@
         "package": "CommonMark",
         "repositoryURL": "https://github.com/SwiftDocOrg/CommonMark.git",
         "state": {
-          "branch": "master",
-          "revision": "00d3a8ce879b6578979a3470430867f85af26b92",
-          "version": null
+          "branch": null,
+          "revision": "62176a332884826997928ba578d9d322c16772fb",
+          "version": "0.4.0"
         }
       },
       {
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/SwiftDocOrg/Markup.git",
         "state": {
           "branch": null,
-          "revision": "9a429d0011d738059bc94f5f92ee406689597a91",
-          "version": "0.0.3"
+          "revision": "029ad8c1115ab32b7c20ab52eb092fbc030deb17",
+          "version": "0.0.4"
         }
       },
       {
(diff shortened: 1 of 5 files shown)


mattt:
@kaishin Curious to get your take on this proposed design. Any concerns about the markup here? (That is, any strong preference for Markdown unordered lists instead of HTML definition lists?)

kaishin:
Nice! I think adding the short description is a huge improvement.

To answer your question, I think we need to consider the different use cases. In general, I try to not introduce HTML tags in Markdown to stay as close to the original vision of maximizing readability as possible. 

If in this case the Markdown is used purely as an intermediate format for GitHub wikis, then introducing HTML isn't such a big deal. But if users are likely to check the markdown documentation into version control, I think it will hurt plain text readability for those who opt to read the docs in a text editor or a terminal.

I often struggle with this back and forth between the readability of the raw source and the semantic correctness of the output, but if I had to make a snap recommendation in this case, I'd go for dropping them in favor of an unordered list.

mattt:
Thanks for your feedback, @kaishin! I just pushed a commit to test out what a list would look like with 443096a35c578c9c2ead52e18e591b47e462c628.

Here's what that looks like in our Alamofire wiki: https://github.com/SwiftDocOrg/Alamofire/wiki/Home/b6a18ab2dc724b6ed6ee0c6150198e6a31490b5b

kaishin:
Looks totally acceptable! I imagined a bit worse in my head 😅

mattt:
Me too! Thanks for the extra set of 👀. Really happy where we ended up with this.

Issue #98, opened 2020-05-05
Exit code 132 when generating HTML


MortenGregersen:
When generating HTML docs the swift-doc command fails with exit code 132, but succeeds when generating markdown.

The Github Actions log can be seen here: [https://github.com/FluxorOrg/Fluxor/pull/57/checks?check_run_id=646042516]()

I tried with the swift-doc Github Action, but it fails with no error, as seen here: [https://github.com/FluxorOrg/Fluxor/runs/640091636]()

Neither way can I get it to output HTML. It works when generating HTML locally on my Mac.

mattt:
Thanks for filing this issue, @MortenGregersen. I'm sorry that this isn't working for you at the moment. Without digging into this too much, it looks like it could be related to #78.

mattt:
This should be fixed by #99.

@MortenGregersen I'm working to cut a new release (1.0.0-beta.3) this week, but you're welcome to try the latest in master in the meantime.

MortenGregersen:
I tried again.

When checking out and building manually I still get the 132 exit code:
https://github.com/FluxorOrg/Fluxor/runs/695438858

And when using the Github Action I get nothing but a failed step:
https://github.com/FluxorOrg/Fluxor/runs/695463951

mattt:
This should be fixed by #140. Holding for verification after Friday's release.

PR #99, opened 2020-05-06
Fix "malloc: Double free of object"

mattt:
Possible resolution for #78 and #98, based on @kean's [observation about the thread unsafety of `Interface`](https://github.com/SwiftDocOrg/swift-doc/issues/78#issuecomment-612307211).


mattt:
I can't reliably reproduce the current crash, but I am reasonably sure about the underlying cause and, anecdotally, haven't encountered this after applying this patch. Tests are passing, and there's no significant change to performance, so I think the best approach would be to merge into master and get more eyes on this in the wild.

PR #100, opened 2020-05-06
Log warning message when failing to generate graph

### LINKED ISSUE #48: "Permission denied" error, documentation is still generated
### Steps to reproduce?

- Project: https://github.com/kaishin/Gifu
- How I installed swift-doc: `git clone` then `make install`
- Exact command: `swift-doc generate ./Sources/ --module-name Gifu --output Documentation --format html`
### Expected behavior

- HTML documentation generated at `./Documentation`
- No error output

### Actual behavior

- HTML documentation generated at `./Documentation`
- The following output
```shell
Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"
Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"
```
### Environment
* OS version: 10.15.4 (19E266)

mattt:
Resolves #48 

As of #52, we log any errors caught during graph generation:

```
2020-05-06T12:00:00-0700 error: Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"
```

Although this involves error handling, the `error` log level is inappropriate here, because the failure isn't critical. Documentation may be generated even if GraphViz graphs aren't available.

This PR changes the log level to `warning` and wraps the raw `NSError` with some more actionable information:

```
2020-05-06T12:00:00-0700 warning: Failed to generate relationship graph for _____________. Please ensure that GraphViz binaries are accessible from your PATH. (Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied")
```

I was able to reproduce #52 locally by running `brew unlink graphviz`. Since [the formula for `swift-doc`](https://github.com/SwiftDocOrg/homebrew-formulae/blob/master/Formula/swift-doc.rb) now includes GraphViz as an (optional) dependency, this is much less likely to occur going forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ as well as top-level type aliases, functions, and variables.
 ## Requirements
 
 - Swift 5.2
+- [GraphViz][graphviz] _(optional)_
 
 ## Command-Line Utility
 
--- a/Sources/swift-doc/Supporting Types/Components/Relationships.swift
+++ b/Sources/swift-doc/Supporting Types/Components/Relationships.swift
@@ -49,7 +49,7 @@ struct Relationships: Component {
         do {
             return try HypertextLiteral.HTML(String(data: graph.render(using: algorithm, to: .svg), encoding: .utf8) ?? "")
         } catch {
-            logger.error("\(error)")
+            logger.warning("Failed to generate relationship graph for \(symbol.id). Please ensure that GraphViz binaries are accessible from your PATH. (\(error))")
             return nil
         }
     }


mattt:
Thanks for reviewing this, @kaishin!

> PS: Not tied to this PR but I think emphasizing that GraphViz is a peer dependency would prevent users from getting confused by this error message. Doesn't make for a good experience to see this error on first run. Either that or skipping this code branch when not running the `diagram` subcommand.

I agree that this is an important detail to get right. Do you have a specific idea for where we can / should communicate `swift-doc`'s relationship to the GraphViz binaries?

To clarify: The GraphViz binary is only used to generate SVG diagrams in the HTML output of the `generate` subcommand. The `diagram` subcommand imports the `GraphViz` Swift module dependency, but doesn't actually require any binaries to generate the DOT language output. (You can do `brew unlink graphviz` and then `swift doc diagram` without any errors).

kaishin:
Thanks for the clarification!

> Do you have a specific idea for where we can / should communicate swift-doc's relationship to the GraphViz binaries?

The first idea I got was to have a "Peer Dependencies" subsection in the readme. If that's too much than a note right below where it's first mentioned could do the job.

A better way would be to let the user know whenever they run the `generate` subcommand, but I am not sure how bullet-proof checking for the executable can be.

mattt:
> The first idea I got was to have a "Peer Dependencies" subsection in the readme. If that's too much then a note right below where it's first mentioned could do the job.

I'm not as familiar with the term "peer dependency". Based on [this write-up](https://nodejs.org/es/blog/npm/peer-dependencies/), I don't think this is quite the right term. But! It would be a great way to describe any generator plugin system we might add in the future. 

For now, I added GraphViz as an "_(optional)_ requirement" (oxymoronic though that may be) with da2cd6c. That seemed like a reasonable first step.

> A better way would be to let the user know whenever they run the `generate` subcommand, but I am not sure how bullet-proof checking for the executable can be.

I think warning on failure is a reasonable solution. It has the advantages of not making extra work and not having the potential for false positive or negative results. Like with validating email addresses, the only real way to see if an email is valid is to send a message and see if they get it.

kaishin:
I agree that the term “peer dependency” isn’t quite right here. I also thought of “soft” and “indirect”.

At any rate, what you proposed is probably enough for now. If it isn’t if won’t be the last time this comes up 😉

mattt:
> At any rate, what you proposed is probably enough for now. If it isn’t if won’t be the last time this comes up 😉

@kaishin Indeed. Thanks for sharing your thoughts about this. All really valuable groundwork as we continue to negotiate and evolve how all the different pieces fit together here.

Issue #105, opened 2020-05-12
Support for @available attributes 


MattKiazyk:
Currently, swift-doc only shows the `@available(...` grouped together with the code 

<img width="830" alt="Screen Shot 2020-05-11 at 10 31 56 PM" src="https://user-images.githubusercontent.com/1119565/81635617-3b474680-93d7-11ea-878f-436a3e63eaf2.png">

Would it be possible for swift-doc to support `@available` attributes, so that end users can see which protocols, functions, etc are no longer supported?

I've come with some Acceptance Criteria of sorts that I think is needed. 

- `@available(...` should be split out from the main code description
- Certain attributes (like `deprecated`) should be shown with the `Deprecated` marker similar to Apple Docs ex: https://developer.apple.com/documentation/uikit/uiactionsheet
- Certain attributes (like `renamed`) should also show a warning section with the string at the top of the view. ex: https://developer.apple.com/documentation/security/secaccesscontrolcreateflags/1617988-touchidany
- Anywhere we show the property and it's`deprecated` we should strikethrough. ex: https://developer.apple.com/documentation/uikit/uiactionsheet

MattKiazyk:
@mattt just wondering if there's anything I can do to get this moving forward in #106 ? In my current project, I have a few deprecated methods which could use this update. More then willing to start updating the output, but just was making sure the direction in the PR is the way you are seeing it?

PR #106, opened 2020-05-12
Add availabilityAttributes property to Symbol

MattKiazyk:
This is to partially support #105 and the `@available` attributes. 

### AvailabilityAttribute
I've added a property to `Symbol` called `availabilityAttributes` to support any attributes that are on the Symbol. `SwiftSemantics` already has support for this, so just drawing out the specific `available` attribute into it's own property

```
 public var availabilityAttributes: [AvailabilityAttribute] {
```

### enum AvailabilityAttributeType

I decided to code all the specific attributes inside the `AvailabilityAttributeType` enum. This allows the HTML or MarkDown, to check for specific attribute and change the UI accordingly. 

The ` init?(from argument: Attribute.Argument) {` is a little awkward I will admit. I was trying to handle the various scenarios of `@available` where sometimes you can have a version, sometimes not. It turned into something a little ugly, but looking for feedback, and perhaps you had another direction for it.

There is no UI change for HTML or Markdown, but the first step in supporting making that change.


mattt:
Thanks for taking this on, @MattKiazyk!

I agree with your assessment in #105 that `swift-doc` should incorporate availability information into its generated output. And adding styling cues, like ~striking out~ deprecated symbols would be a quick win. Eventually, it'd be nice to reconcile individual symbol availability with the package's declared deployment targets.

I keep going back and forth about whether this functionality belongs in `SwiftDoc` or [SwiftSemantics](https://github.com/swiftdocorg/swiftsemantics). But for now, I think  the right approach would be have an `Availability` type that initializes from a `SwiftSemantics` `Attribute` value, as well as eventually from a Swift package manifest, compiled artifact, or index entry. That way, we can support both the short- and long-term goals stated above.

> The ` init?(from argument: Attribute.Argument) {` is a little awkward I will admit. I was trying to handle the various scenarios of `@available` where sometimes you can have a version, sometimes not. It turned into something a little ugly, but looking for feedback, and perhaps you had another direction for it.

Some of that awkwardness is a consequence of `@available` being a shared spelling for two separate concepts: language availability and platform availability, with the latter having both a longhand and shorthand form. I wrote more about this in [this NSHipster article](https://nshipster.com/available/).

What I'd like to see from this PR is this:

- An `Available` type with better alignment with how the Swift compiler / Clang understands API availability
- Support for arbitrary platform names, both to accommodate existing ones like `_PackageDescription` and `Linux`, as well as future ones like `wasm` and `Windows`.
- Comprehensive tests for initializing from availability attributes. Given the wide variety of possible inputs, we should do our best to provide extensive coverage. For reference, take a look at [the Swift compiler's tests for availability versions](https://github.com/apple/swift/blob/master/test/Sema/availability_versions.swift)

mattt (changes requested):
Thanks for taking this on, @MattKiazyk!

I agree with your assessment in #105 that `swift-doc` should incorporate availability information into its generated output. And adding styling cues, like ~striking out~ deprecated symbols would be a quick win. Eventually, it'd be nice to reconcile individual symbol availability with the package's declared deployment targets.

I keep going back and forth about whether this functionality belongs in `SwiftDoc` or [SwiftSemantics](https://github.com/swiftdocorg/swiftsemantics). But for now, I think  the right approach would be have an `Availability` type that initializes from a `SwiftSemantics` `Attribute` value, as well as eventually from a Swift package manifest, compiled artifact, or index entry. That way, we can support both the short- and long-term goals stated above.

> The ` init?(from argument: Attribute.Argument) {` is a little awkward I will admit. I was trying to handle the various scenarios of `@available` where sometimes you can have a version, sometimes not. It turned into something a little ugly, but looking for feedback, and perhaps you had another direction for it.

Some of that awkwardness is a consequence of `@available` being a shared spelling for two separate concepts: language availability and platform availability, with the latter having both a longhand and shorthand form. I wrote more about this in [this NSHipster article](https://nshipster.com/available/).

What I'd like to see from this PR is this:

- An `Available` type with better alignment with how the Swift compiler / Clang understands API availability
- Support for arbitrary platform names, both to accommodate existing ones like `_PackageDescription` and `Linux`, as well as future ones like `wasm` and `Windows`.
- Comprehensive tests for initializing from availability attributes. Given the wide variety of possible inputs, we should do our best to provide extensive coverage. For reference, take a look at [the Swift compiler's tests for availability versions](https://github.com/apple/swift/blob/master/test/Sema/availability_versions.swift)

MattKiazyk:
@mattt made some more changes to the `Availability` type. This should better allow arbitrary platform names. 

I tried to best split out the platform/version for future use where we could filter based on that.  

Also added more tests for a variety of different `@available` options.

MattKiazyk:
@mattt just wondering if there's anything I can do to move this forward?

PR #114, opened 2020-05-17
Fix indentation on code blocks inside discussion elements

samsymons:
This PR fixes up an issue where code blocks were receiving inconsistent indentation.

The root cause here comes from the discussion in https://github.com/SwiftDocOrg/swift-doc/issues/113, where the `text-indent` CSS property was [being applied to all code blocks](https://github.com/SwiftDocOrg/swift-doc/blob/master/Assets/css/all.css#L887), rather than just those outside of discussion elements.

My guess here is that `text-indent` should stay on all code blocks _except_ those with the `.discussion` class, where it's not uncommon to have multi-line code blocks, so this PR sets `text-indent` to 0 and adjusts the padding accordingly to remove the `padding-left`.

I'm still learning the internals of SwiftDoc and currently dusting off years-old CSS knowledge, so give me a shout if this solution is off base! ❤️ 

---

Fixes https://github.com/SwiftDocOrg/swift-doc/issues/113.


mattt:
Thanks for taking this on, @samsymons! At first glance, this looks right. I'll have a chance to try it out locally in the next day or two.

mattt:
😅 Whoops, I neglected to add a changelog entry for this. Added that with 5dafcb1.

Thanks again for your help on this, @samsymons!

PR #116, opened 2020-05-20
Generate docs only for public symbols

ApolloZhu:
## Changes

- Renders only symbols that are public
- To fix problem with protocol conformances not showing up, now assumes Unknown isPublic. This is a little shaky, but if we can use it without knowing the source, it should be public, right?

---

## Before

image

### # Where private declarations are shown:

image

### # And because `Node` is private, when clicking on it, you'll get:

image

## After

image

<details>
<summary>Full TOC now don't have private members</summary>
 
image

</details>


ApolloZhu:
Another example of beta3 rendering private members:
https://sindresorhus.com/playground/KeyboardShortcuts_Shortcut/#keyboardshortcuts.shortcut.keytocharacter()

mattt:
Sorry for not getting to this sooner, @ApolloZhu. This is an excellent PR, and am very happy to get this into the forthcoming beta release. Thank you!

PR #125, opened 2020-05-22
Added Mint installation instructions to README.md

jimmya:


mattt:
Thanks for the PR, @jimmya.

I'm not very familiar with Mint. According to the README, this is how it positions itself:

> Homebrew is a popular method of distributing Swift executables, but that requires creating a formula and then maintaining that formula. Running specific versions of homebrew installations can also be tricky as only one global version is installed at any one time. Mint installs your package via SPM and lets you run multiple versions of that package, which are installed and cached in a central place.

Since we already support Homebrew, what reason would we have to support Mint as well?

jimmya:
Hi Mattt,

The thing I like about Mint is that it’s written in Swift obviously ;).

But on a more serious note. It allows you to define a Mintfile in which you can specify al your swift tools and pin them to specific versions to ensure consistency across contributors and in CI. So for instance you can asure a specific version of SwiftLint is used across your team. 

With Mint all I need to setup CI wise for my tooling is to install Mint once. Moving on I can simply run ‘mint bootstrap’ and I have confidence all my tools work as expected. Without having to deal with multiple sources of my tools.

Also the project currently works with Mint as there is a executable defined in the Package.swift file. This addition to the readme is pure informational. 

But surely I can understand if this addition might not fit into the ideology behind this project. I’ll happily leave that up to you of course.

Best regards,

Jimmy

mattt:
Thanks for taking the time to answer my question, @jimmya. 

Right now, I don't really see a compelling reason to support Mint. If it happens to work now, that's great. But I can't guarantee that we won't, in the future, change the `Makefile` in a way that can't be handled by Mint (I've had to do this in the past to link libSyntax correctly). 

Candidly, I still don't quite understand the appeal from your perspective... For your use case specifically, is there any advantage of a `Mintfile` / `mint bootstrap` over [`Brewfile` / `brew bundle`](https://github.com/Homebrew/homebrew-bundle)? How do you manage non-Swift dependencies? I'm genuinely curious.

jimmya:
I admit Mint isn't a silver bullet by a long shot yet. My inner Swift enthusiast always strives to limit my use of non swift tools where possible. But yeah for now I handle non Swift dependencies with custom scripting.

But I hope in the future, with binary dependency support in SPM this would be achievable.

But thanks for your time and I fully understand your rationale to decline this PR.

mattt:
> I admit Mint isn't a silver bullet by a long shot yet.

I don't mean for this to be pedantic or confrontational, but I think it's important to note that when we talk about "silver bullets" in software, it's usually a reference to Fred Brooks' essay _["No Silver Bullet"](https://en.wikipedia.org/wiki/No_Silver_Bullet)_: 

> There is no single development, in either technology or management technique, which by itself promises even one order of magnitude improvement within a decade in productivity, in reliability, in simplicity.

The point is that there are _no_ silver bullets. To the point that it applies in this context (which I'm not 100% on), neither Mint nor Homebrew are or will ever be silver bullets.

> My inner Swift enthusiast always strives to limit my use of non swift tools where possible. But yeah for now I handle non Swift dependencies with custom scripting.

Don't get me wrong: I like Swift. I like it so much that I decided to work at Apple for the sole reason that I could help make Swift better. At the same time, I think it's really important to try to maintain objectivity when we make technical decisions. There are practical reasons for maximizing the use of a single programming language in an application / ecosystem, but those should be kept separate from emotional reasons. 

Right now, Homebrew effectively solves software distribution on macOS. That's not to say that it's an ideal solution (it's not) or that can't be improved upon (it can) or that there isn't room for new solutions ([Nix](https://nixos.org/nix/) looks nice). But I don't fully understand the value proposition of Mint, except as an exercise in dog-fooding Swift.

For any project, you have a small budget of ["innovation tokens"](https://mcfunley.com/choose-boring-technology). How you decide to invest these tokens is a strong predictor of how successful you'll be at solving the problem at hand.

Issue #126, opened 2020-05-24
Relative links are broken in markdown


mattpolzin:
When generating markdown, especially for use in GitHub, I like to rely on relative links as a robust way to link between documentation files living at file paths consistent relative to each other.

When I run the current GitHub Action, the default `baseURL` is used ("/"). This results in broken links between documentation files within GitHub or the GitHub Wiki. 

I think this is partially solved by exposing the `--base-url` argument via the GitHub action (should this also default to "./" to coincide with the default format of `.commonmark`?).

That is only half the solution, though, because setting the `baseURL` to "./" results in `Links` prefixed with "/./". The solution to the second part of this _appears_ to be straight forward enough, but time, experience, and feedback from maintainers will prove or disprove that!

mattpolzin:
I've got this WIP. Will PR soon.

PR #127, opened 2020-05-24
Bugfix/relative links

mattpolzin:
Fixes https://github.com/SwiftDocOrg/swift-doc/issues/126

> When generating markdown, especially for use in GitHub, I like to rely on relative links as a robust way to link between documentation files living at file paths consistent relative to each other.
> 
> When I run the current GitHub Action, the default baseURL is used ("/"). This results in broken links between documentation files within GitHub or the GitHub Wiki.
> 
> I think this is partially solved by exposing the --base-url argument via the GitHub action (should this also default to "./" to coincide with the default format of .commonmark?).
> 
> That is only half the solution, though, because setting the baseURL to "./" results in Links prefixed with "/./". The solution to the second part of this appears to be straight forward enough, but time, experience, and feedback from maintainers will prove or disprove that!

This PR
1. Exposes the `base-url` argument via the GitHub action. Defaults the GitHub action's `base-url` to "./". I think a relative URL works well more often than not for markdown, but I am not at all tied to this default because it also makes sense to use the same default the underlying program uses: "/".
2. Uses `URL` for path append operation. Ostensibly, this simplifies the code and fixes my problem, but I don't know for sure that this works for all use-cases; I would love feedback on this decision.

I ran documentation generation with a few believable base urls to check some common cases. For a component named `"A.B"`:
1. "" -> `[A.B](A_B)`
2. "/" -> `[A.B](/A_B)`
3. "something" -> `[A.B](something/A_B)`
4. "/something" -> `[A.B](/something/A_B)`
5. "https://website.com/docs" -> `[A.B](https://website.com/docs/A_B)`
6. "https://website.com/docs/" -> `[A.B](https://website.com/docs/A_B)`
7. “./“ -> `[A.B](./A_B)`


kongzii:
It would be great to get this merged. Currently, I have to fix links at Home.md after every auto-generation.

mattpolzin:
If anyone wants to temporarily switch to a fork, this fix has been working well for me for a while now. Just update your GitHub job as follows and you can change it back when the problem is fixed in this repo.

```yaml
  ...
      - name: Generate Documentation
        uses: mattpolzin/swift-doc@bugfix/relative-links
  ...
```

mattt:
@mattpolzin Thanks for submitting this, and sorry for not responding sooner. I'm taking a look now.

Sources/swift-doc/Supporting Types/Page.swift
```diff
@@ -70,11 +70,7 @@ func writeFile(_ data: Data, to url: URL) throws {
...
 fileprivate extension URLComponents {
     func appendingPathComponent(_ component: String) -> URLComponents? {
-        var urlComponents = self
-        var pathComponents = urlComponents.path.split(separator: "/").map { "\($0)" }
-        pathComponents.append(component)
-        urlComponents.path = "/" + pathComponents.joined(separator: "/")
-
-        return urlComponents
+        return url.map { $0.appendingPathComponent(component) }
+            .flatMap { URLComponents(url: $0, resolvingAgainstBaseURL: true) }
```
mattt, line 74: I slightly prefer the solution in #139 (though I need to test it out a bit more to make sure they're equivalent). I think we can all agree though that my code here, in addition to being buggy, is totally bizarre and anything would be an improvement.

action.yml
```diff
@@ -12,6 +12,10 @@ inputs:
   format:
     description: "The output format (commonmark or html)"
     default: "commonmark"
+  base-url:
+    description: "The base URL for all relative URLs generated in documents"
+    required: false
+    default: "./"
```
mattt, line 18: Regardless of what happens with `Page.swift`, I'm definitely happy to have `base-url` exposed in the action interface.

Sources/swift-doc/Supporting Types/Page.swift
mattpolzin: I've been staring at the two implementations for a few minutes now. Ultimately I wish there were unit tests to prove it, but I like that #139 does away with the `URLComponents` abstraction entirely. It sure looks to me like the implementation in #139 introduces a bug where all paths become relative, though -- note that it takes the `.path` of the URL instead of the `.absoluteString`.

I think we can have that implementation if `.path` is swapped for `.absoluteString` and everything will be better all around. I proved this to myself in a Playground, at least.

```swift
let baseURLs = [
    "https://website.com",
    "./",
    "/"
]

let identifierStandin = "identifier_standin"

let mattsURLs = baseURLs.map { URLComponents(string: $0)?.url.flatMap { URLComponents(url: $0.appendingPathComponent(identifierStandin), resolvingAgainstBaseURL: true)?.string } }
let karesURLs = baseURLs.map { URL(string: $0)?.appendingPathComponent(identifierStandin).path }

let bestOfBothWorlsURLs = baseURLs.map { URL(string: $0)?.appendingPathComponent(identifierStandin).absoluteString }

func pretty(_ arr: [String?]) -> String {
    arr.map { $0 ?? "nil" }.joined(separator: "\n")
}

print(pretty(mattsURLs))
print("====")
print(pretty(karesURLs))
print("====")
print(pretty(bestOfBothWorlsURLs))
```

```
https://website.com/identifier_standin
./identifier_standin
/identifier_standin
====
/identifier_standin
./identifier_standin
/identifier_standin
====
https://website.com/identifier_standin
./identifier_standin
/identifier_standin
```

mattpolzin: I merged with #139 and then made the aforementioned change from `.path` to `.absoluteString`. Should make it a clean merge whether you merge #139 before merging this or not.

mattt:
@mattpolzin @kareman Thank you both for your help in identifying and fixing this bug, and for your patience as I got this merged. I look forward to cutting a new release with this and other improvements this Friday.

PR #131, opened 2020-06-03
Regression - private subclasses showing

MattKiazyk:
Issue: Private symbols are showing up in pages when they inherit from a public symbol. 

I believe this is a regression caused by the refactor in https://github.com/SwiftDocOrg/swift-doc/pull/99. This was working correctly in 1.0-beta.2

The issue is that `symbols` is being used in `relationship` closure, but instead of using the filtered var, it's using the symbol property that's passed into the init - which is not filtered.  This causes private subclasses to show on a parent page. The private page wasn't created, causing a missing link. 

I went with the easy filter inside the closure option fix. 

I also wrote a quick test to hopefully stop any reggressions in the future where any non-public subclasses should never be shown anywhere.


mattt:
Hey @MattKiazyk, sorry for not responding sooner. Thank you for submitting this PR! Everything is looking great, and I'm excited to include this as part of Friday's forthcoming release.

PR #132, opened 2020-06-06
Only show scrollbars when needed

andrewchang-bird:
Changes `overflow: scroll` to `overflow: auto` so that scroll bars only appear if the content is scrollable. This primarily affects macOS users who opt to always show scroll bars for accessibility reasons (and users on platforms that don't auto-hide scroll bars).

Currently, the docs look like this (notice the scroll bars for non-scrollable content in the nav and code blocks):

Old

This change hides scroll bars for non-scrollable content, which seems like the expected behavior:

New non-scrollable

Content that is actually scrollable still shows scroll bars:

New scrollable


mattt:
Hey, @andrewchang-bird! Thank you for taking the time to write this up and submit a patch. I just merged it, and look forward to cutting a new release on Friday that includes this fix.

Issue #101, opened 2020-05-07
Add the ability to hide certain properties/methods


sindresorhus:
When you subclass a class, you sometimes have to implement some required methods or you want to override a property. These have to be marked as `public`, and hence shows up in the docs. It would be nice `swift-doc` could have a special comment or something to hide it.

When using TypeScript and `type-doc`, I can do something like this:

mattt:
Thanks for sharing this feedback, @sindresorhus. It's certainly possible to add an annotation to hide declarations from `swift-doc` — and that may well be the solution go with — but I'd like to explore the motivating use cases to see if there might be a better solution.

Attempting to restate the problem: 

* Some public-facing APIs are effectively implementation details of type identity, and aren't typically documented. 
* Undocumented symbols in generated output are conspicuous, or at least compete with more important information.
* These can be divided into two cases<sup>*</sup>: 

  1. Requirements of an adopted protocol. For example, the [`description` property requirement for `CustomStringConvertible`](https://developer.apple.com/documentation/swift/customstringconvertible/1539130-description).
  2. Overridden members that are inherited from a superclass. For example, the [`canBecomeKeyView ` property of `NSView`](https://developer.apple.com/documentation/appkit/nsview/1483759-canbecomekeyview).

<sup>*</sup> _Are there any others you can think of?_

* * *

In the first case, there's nothing to be added from a documentation comment; the behavior is defined by the protocol. 

Here, I think the best solution would be to do the following: 

1. Inherit any documentation from the protocol declaration
2. Organize the symbol into a group with others for that protocol that are separate from type-specific declarations.

<br/>

In the second case, it's less clear when a documentation comment adds useful information. Sometimes, the reason for a method being overridden is an implementation detail that shouldn't be documented. Other times, such as the example of `canBecomeKeyView`, the overridden implementation may completely change the original behavior; this would probably benefit from documentation.

I'm less certain about what a good solution would be for this case. 

Hiding the declaration is one option, but I worry that doing so can be counter-productive, as the fact that a member is overridden by a subclass may be valuable information to an API consumer attempting to understand a certain behavior. 

What do you think about a similar approach to the one proposed for the first case? What if we had a section for `Overridden Members` that was subdivided according to the original declaration. For example, `class MyView: NSView {...}` would have `canBecomeKeyView` under `NSView` and `acceptsFirstResponder` under `NSResponder`.

sindresorhus:
[...]

Issue #107, opened 2020-05-12
Group enum cases based on `MARK` comments


sindresorhus:
For https://github.com/sindresorhus/KeyboardShortcuts/blob/49b017b2450ac37353934d94d6c64957b62c45b4/Sources/KeyboardShortcuts/Key.swift#L8-L34, I get the following: https://sindresorhus.com/playground/KeyboardShortcuts_Key/

With Jazzy, I get the enum cases grouped too: https://sindresorhus.com/KeyboardShortcuts/Classes/KeyboardShortcuts/Key.html#/Letters

mattt:
@sindresorhus Curious to know, do you have any thoughts about how `MARK` headings should relate to declaration type groupings? Should we treat them as peers / alternates or should we see the current "by type" grouping just as the default sections for when `MARK` headings aren't provided?

sindresorhus:
My preference would be for them to be sub-headings under the `Enumeration Cases` section. Jazzy makes them top-level sections, which feels wrong to me.

mattt:
@sindresorhus What if a section contains symbols of more than one type? For example:

```swift
enum E {

// MARK: Example

case c
func f() {}
}
```

sindresorhus:
Not sure, to be honest. I wouldn't use `MARK` like that.

PR #142, opened 2020-07-11
Ability to exclude symbols from generated documentation

biocross:
While a discussion is underway on how to exclude certain methods and properties in #101, this PR adds an argument to the `generate` command to add the ability to exclude full top level symbols from the generated documentation.

This is done using an external file containing a list of line separated symbol names, like:

```
MyExcludedClass
MyMockClass
```

This PR:

- Adds an `excluded-symbols` argument in the `generate` command options that takes in a path to a file containing a line separated list of symbols to be excluded. Defaults to `nil`.
- Adds a new argument to the `Module` initialiser, taking in an array of `exclusions` and filters the symbols by them.
- Updates `README` and `action.yml` with the new argument.


mattt:
Thanks for sharing this PR, @biocross. I really appreciate your taking the time to work on this.

I'm glad to see an alternative strategy to controlling the output of symbols in documentation (as opposed to a directive like `@hidden` or `:nodoc:`). However, I'm still not entirely sold on the idea in the first place... Have you needed something like this in your own projects? I'd be interested to hear examples of when a `public` API shouldn't be documented.

biocross:
Hi @mattt!

For my use case, it's mostly been the documented symbols competing with more important information. Since there's no way to currently **manually section** / **order** the generated documentation instead of the default sectioning by symbol type (enums / structs / protocol), I decided to have an exclusion list to keep important information at the top..

mattt:
I appreciate your expanding on that, @biocross. There's a discussion about how section headers / symbol organization should work in #107, and I'd be interested to get your take on that.

> An example was to hide some unit testing related mock symbols, which ideally should be documented and included, but were being put at the top of generated documentation.

Curious to know more about this specific case. I wouldn't normally consider test-related symbols to be part of a public API (and therefore be documented) unless testing were the sole purpose of the module. Is there any reason why those mocks couldn't be `internal` and imported with the `@testable` attribute?

karwa:
>  However, I'm still not entirely sold on the idea in the first place... Have you needed something like this in your own projects? I'd be interested to hear examples of when a public API shouldn't be documented.

I was looking for this feature when I came across this. I could describe a couple of use-cases I have, if it helps:

- Types exposed via `typealias`es.

  I have a lazy collection type, `LazilyPercentEncodedGroups`, which does percent-encoding as a 2-dimensional collection - the `Element` of the collection is itself another collection, consisting of either the 3 bytes of the percent-encoded codepoint, or the original codepoint if it was not encoded. This is a useful type for my internal use, but it's not so useful as a public API. Users should really call `.joined()` on this to collapse the 2D collection. So, for the public API, I've defined a `typealias LazilyPercentEncodedUTF8<Source> = FlattenSequence<LazilyPercentEncodedGroups<Source>>`.

  `LazilyPercentEncodedGroups` has to be public because it appears in a public typealias, but it isn't intended to be used directly. The constructor is internal and there is no way AFAIK to dig the underlying instance out from the `FlattenSequence`, meaning users should never encounter it.

- Associated types of protocol conformances for said types.

  As mentioned, the `Element` of `LazilyPercentEncodedGroups` is a collection called `_PercentEncodedByte`. As `LazilyPercentEncodedGroups` is public, the `Collection` conformance and associated `Element` type must also be made public.

  Similarly, I have a `LazilyPercentDecoded` type (which is public and intended to be used directly), with a custom, opaque `Index`. `swift-doc` lists the `Index` type separately, which clutters the type list. There isn't really anything anybody can do with the `Index` - it conforms to `Equatable` and `Comparable` as `Collection` requires, but has no other API.

  image

  The same goes for custom `Iterator` types:
  
  image

- Funky API design!

  Swift libraries often have exotic API designs in order to make their usage sites clearer and guide features such as generics and type-inference. I have an example in the same library with a set of types which should preferably be referred-to by KeyPath as a guide for type-inference!

Issue #129, opened 2020-05-27
Code blocks are not indented


mbrandonw:
The following documentation:

https://github.com/pointfreeco/swift-composable-architecture/blob/6483439efdcf3d24b8ad6e9dea73fbe64c0531d2/Sources/ComposableArchitecture/Reducer.swift#L30-L46

Is rendered like so:

<details>
<summary>Expand</summary>
<img width="853" alt="image" src="https://user-images.githubusercontent.com/135203/83058541-618ef800-a00d-11ea-8019-178b7fc225af.png">
</details>

Dropping the `white-space: pre-line` [here](https://github.com/SwiftDocOrg/swift-doc/blob/a2013b64ffbcf91be45e6ecff9e88fdf8019a08d/Assets/css/all.css#L890) fixes it for me, but not sure if there are other implications to think about:

https://github.com/SwiftDocOrg/swift-doc/blob/a2013b64ffbcf91be45e6ecff9e88fdf8019a08d/Assets/css/all.css#L890

mattt:
Ah, this looks like a regression related to #114.

mattt:
Fixed by #144. Thanks again for your help with this, @mbrandonw!

PR #144, opened 2020-07-25
Use pre instead of pre-line for code whitespace.

mbrandonw:
This fixes the issue I filed https://github.com/SwiftDocOrg/swift-doc/issues/129

I think you want to use `white-space: pre` here in order to preserve spaces. `pre-line` will collapse the spaces:

> Sequences of whitespace will collapse into a single whitespace. Text will wrap when necessary, and on line breaks

Another option is `pre-wrap` if you want long lines to wrap.


mattt:
This looks great, @mbrandonw. Thank you for submitting this PR and for your patience as I ramp back up to working on swift-doc on a more regular basis. I look forward to cutting a new release by Friday with this fix and a whole raft of improvements.

mbrandonw:
@mattt that is awesome to hear and no worries! really excited for the future of this project and we will be adding it to TCA shortly!

PR #146, opened 2020-07-28
FIX: swift-doc creates file and directory with unexpected permission

niw:
**Problem**

swift-doc creates file with execution bit, also creates directory
without execution bit.

**Solution**

Remove unexpected permission attributes given to the `FileManager`
and let them to use default permission respects file system and process umask.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -31,6 +31,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   #131 by @MattKiazyk.
 - Fixed display of bullet list items in documentation discussion parts.
   #130 by @mattt.
+- Fixed file and directory unexpected permissions.
+  #146 by @niw.
 
 ## [1.0.0-beta.3] - 2020-05-19
 
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -50,7 +50,7 @@ extension SwiftDoc {
       let baseURL = options.baseURL
 
       let outputDirectoryURL = URL(fileURLWithPath: options.output)
-      try fileManager.createDirectory(at: outputDirectoryURL, withIntermediateDirectories: true, attributes: fileAttributes)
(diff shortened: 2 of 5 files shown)


mattt:
Thanks for the patch, @niw! That was some code left over from the initial proof-of-concept. For whatever reason, I thought I needed to override permissions to get it working on CI.

Could you please add a corresponding entry into `Changelog.md`? As soon as we have that, this will be ready to merge.

niw:
> I thought I needed to override permissions to get it working on CI.

So, while working on this patch, I was investigating corelibs Foundation and found the next behavior (possibly a bug.)
No updates yet from Swift team, but this could be a potential reason why setting the permissions was needed for CI, which I think might run with corelibs Foundation.

https://bugs.swift.org/browse/SR-13307
(Seems like caused by apple/swift-corelibs-foundation#1876)

mluisbrown:
Looks like swift-doc is still creating files with unexpected permissions, which means an extra `'sudo chown --recursive $USER Documentation'` step is needed in a GitHub workflow before being able to use https://github.com/JamesIves/github-pages-deploy-action to deploy documentation.

Without this extra step, I get "Permission Denied" when trying to copy the documentation files: https://github.com/JamesIves/github-pages-deploy-action/issues/388

niw:
@mluisbrown Hi, do you know if it’s because the files generated by swift-doc has no read permissions from the others?
e.g. I am wondering if `$USER` that sudo command taken is the same as the user who runs swift-doc on your CI.

Because that is slightly separated issue but actually Swift corelibs Foundation bug (I would say, it’s bug,) as I describe before on this thread caused by apple/swift-corelibs-foundation#1876.

Also if that’s the case, when you can run the CI on macOS instead of other POSIX system such as Linux, then probably the result would be different, which is using Apple's Foundation not corelibs one.

mluisbrown:
@niw I don't actually know. It's likely that the problem is the Foundation bug that you mentioned, and not anything specifically related to swift-doc itself 👍

niw:
FYI here, I just posted a potential fix on apple/swift-corelibs-foundation#2851 for the permission issue.

PR #178, opened 2020-09-04
Don't render href on nodes for unknown symbols for child relationships

MattKiazyk:
When rendering the GraphViz diagram, when a symbol has a relationship of an `Unknown` type, the code was still rendering those with `href` attributes, causing links be created to invalid links.

A check was added on the base node for `Unknown` types, however no check was added to the relationship loop. This PR adds that.

**Before:** 

Screen Shot 2020-09-04 at 9 22 49 AM

**After: **
Screen Shot 2020-09-04 at 9 13 15 AM


mattt (approved):

mattt:
Great work, @MattKiazyk. Thanks so much for this. I'll merge this in as soon as CI finishes.

Issue #111, opened 2020-05-12
Feature request: publish prebuilt Docker image to Docker Hub to speed up the GitHub Action


hallee:
Hi, thanks for making this available as a GitHub Action!

Unfortunately it has to run the `builder` steps of your Dockerfile on each run, which can take 7+ minutes on my GitHub Action instances.

GitHub Actions supports prebuilt Docker images and is much faster using them. If you had a prebuilt Docker image on [Docker Hub](https://hub.docker.com) this would probably run in a few seconds. 

https://help.github.com/en/actions/building-actions/metadata-syntax-for-github-actions#runsimage

> The value can be the Docker base image name, a local Dockerfile in your repository, or a public image in Docker Hub 

I'd open a PR to help with this, but seems like you'll need to set up a Docker Hub user for this project so there's not much I can do.

mattt:
@hallee This is definitely on our radar. Our preference was to use [GitHub Packages](https://github.com/features/packages) as the primary host for pre-built Docker images. Curious to know, do you see any benefit in also publishing to Docker Hub?

hallee:
I think Actions currently only supports remote images from Docker Hub, and not GitHub Package Registry (yet). So the only benefit of Docker Hub is you can do that now instead of waiting for GitHub Actions to support their own Docker registry.

mattt:
Ah, thanks for pointing that out. The [documentation](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainerimage) mentions a  "public docker registry name" (which I assumed included GitHub Packages), but I wasn't able to find any examples of that in the wild.

I have a Docker Hub organization and repository set up. I'd be very happy to get your help in setting up automatic publishing of nightly and stable (tagged) images. What's your Docker Hub username?

LucasCarioca:
Any progress on this? I removed action from my project because it takes way too long.

mattt:
@hallee Now that beta 4 is out, I'll work to get an official Docker image released and hook that up to the GitHub action (which, for better or for worse, tracks master rather than another branch or a tag).

MaxDesiatov:
I'd be happy to help if there's still anything left to do here?

mattt:
@MaxDesiatov Thanks for putting that Docker image together. I just added you as a collaborator to the swiftdoc org on Docker Hub, so you should be able to publish to https://hub.docker.com/repository/docker/swiftdoc/swift-doc.

PR #180, opened 2020-09-14
Add graphviz in Dockerfile

MaxDesiatov:
`graphviz` is not currently installed, which triggers warnings when running the corresponding GitHub Action.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,9 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed missing GraphViz dependency in Dockerfile.
+  #180 by @MaxDesiatov.
 - Fixed listing of function parameters, when generating CommonMark documentation.
   #170 by @domcorvasce.
-
 - Fixed version number for swift-doc command.
   #159 by @mattt.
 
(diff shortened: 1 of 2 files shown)


mattt (approved):

mattt:
Thanks for the PR, @MaxDesiatov. I would have sworn that this was already in there, but I'm probably conflating adding GraphViz as a dependency in the Homebrew formula. Anyway, glad to have this fixed, and looking forward to having this available in the forthcoming release.

MaxDesiatov:
Thanks, I also have a prebuilt image here that I'm using in the meantime to make the action run faster, just in case you're interested https://hub.docker.com/r/maxdesiatov/swift-doc

I hope you could upload the official one in your Docker Hub namespace at some point.

MaxDesiatov:
(This required me to create two separate Dockerfiles in my swift-doc fork, one for the prebuilt image, the other for the action that only references the prebuilt image and does nothing else)

mattt:
@MaxDesiatov

> Thanks, I also have a prebuilt image here that I'm using in the meantime to make the action run faster, just in case you're interested https://hub.docker.com/r/maxdesiatov/swift-doc
> 
> I hope you could upload the official one in your Docker Hub namespace at some point.

That's great! Let's continue this discussion on #111.

PR #183, opened 2020-09-18
Test with Swift 5.3 on Linux

MaxDesiatov:
I'm not sure if you'd like to test with both Swift 5.2 and 5.3 on Linux. As you're using a cache, I think it would be easier to test just with a single version, as it doesn't make sense to share that cache. Let me know if you'd prefer to test both versions after all with separate cache keys.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,8 +27,12 @@ jobs:
   linux:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        swift: ["5.3", "5.2"]
+
     container:
-      image: swift:5.3
+      image: swift:${{ matrix.swift }}
 
     steps:
       - name: Checkout
@@ -36,9 +40,9 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: .build
-          key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
+          key: ${{ runner.os }}-spm-swift-${{ matrix.swift }}-${{ hashFiles('**/Package.resolved') }}
           restore-keys: |
-            ${{ runner.os }}-spm-
+            ${{ runner.os }}-spm-swift-${{ matrix.swift }}-
       - name: Install System Dependencies
         run: |
           apt-get update
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,13 @@ on:
 
 jobs:
   macos:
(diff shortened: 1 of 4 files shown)


MaxDesiatov:
Of course, `SwiftSyntax parser library isn't compatible` 😞

MaxDesiatov:
This requires changes to `swift-doc` dependencies, I'll update those first...

.github/workflows/ci.yml
```diff
@@ -10,35 +10,43 @@ jobs:
...
         env:
-          DEVELOPER_DIR: /Applications/Xcode_11.4.app/Contents/Developer
+          DEVELOPER_DIR: /Applications/Xcode_${{ matrix.xcode }}.app/Contents/Developer
 
   linux:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        swift: ["5.3", "5.2"]
```
MaxDesiatov, line 36: I think this won't work, we'll need two different versions of SwiftSyntax used for each Swift version. The fact that SwiftSyntax can't be linked in statically (or redistributed with `swift-doc` itself) is a deal-breaker 😑

mattt:
@MaxDesiatov Thanks for getting a head start on this! A lot of the work I'd been wanting to do on `swift-doc` had been blocked on general availability of Swift 5.3, so I'm really excited for this to finally be coming in place.

I just pushed a few commits to test Swift 5.2 and 5.3 on CI for Linux and macOS (I'd like to keep support for 5.2 in the short term, to give folks some time to upgrade).

.github/workflows/ci.yml
mattt: We should be able to work around that, no? Could we use Swift version-specific manifests like `Package@swift-5.2.swift`?

MaxDesiatov:
Thanks! This now depends on https://github.com/NSHipster/SwiftSyntaxHighlighter/pull/8

.github/workflows/ci.yml
MaxDesiatov: Spot on, I forgot about that trick, I'll add it to NSHipster/SwiftSyntaxHighlighter#8.

mattt (approved):

mattt:
Thanks for your help with this, @MaxDesiatov! Glad to have this blocker out of the way so that we can target Swift 5.3 for future releases.

MaxDesiatov:
Thanks for finishing the PR!

PR #185, opened 2020-09-24
Use prebuilt Docker image

### LINKED ISSUE #111: Feature request: publish prebuilt Docker image to Docker Hub to speed up the GitHub Action

mattt:
Resolves #111

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed relationship diagram to prevent linking to unknown symbols.
   #178 by @MattKiazyk.
 
+### Changed
+
+- Changed GitHub Action to use prebuilt Docker image.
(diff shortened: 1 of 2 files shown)


MaxDesiatov (approved):
That's awesome! I didn't know about this trick, though I certainly saw `docker://` URL scheme used somewhere before...

PR #193, opened 2020-09-29
Trigger publish from release workflow

mattt:
For the release of [1.0.0-beta.5](https://github.com/SwiftDocOrg/swift-doc/releases/tag/1.0.0-beta.5), I was excited to test out the automated release workflows put in place by #190. Unfortunately, things didn't work exactly as expected. Although a release was created in response to pushing a tag, the deploy workflow didn't trigger in response to that release being created. At first I thought it was an issue with the event type, but then I saw this [in the GitHub Actions documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token): 

> ## Triggering new workflows using a personal access token
>
> When you use the repository's `GITHUB_TOKEN` to perform tasks on behalf of the GitHub Actions app, events triggered by the `GITHUB_TOKEN` will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's `GITHUB_TOKEN`, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.
>
> If you would like to trigger a workflow from a workflow run, you can trigger the event using a personal access token. You'll need to create a personal access token and store it as a secret. To minimize your GitHub Actions usage costs, ensure that you don't create recursive or unintended workflow runs.

This PR passes the `GH_PERSONAL_ACCESS_TOKEN` as `GITHUB_TOKEN` for the `actions/create-release` action, which should allow one workflow to trigger the other.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,83 +5,47 @@ on:
     types: [created, published]
 
 jobs:
-  macos:
-    name: Upload macOS Catalina release binary
-    runs-on: macos-10.15
-
-    strategy:
-      matrix:
-        xcode:
-          - "12" # Swift 5.3
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-        with:
-          ref: ${{ github.event.release.tag_name }}
-      - name: Build and Package
-        run: |
-          make swift-doc
-          tar -cf swift-doc-${{ github.event.release.tag_name }}.catalina.bottle.tar swift-doc
-          gzip -f swift-doc-${{ github.event.release.tag_name }}.catalina.bottle.tar
-        env:
-          DEVELOPER_DIR: /Applications/Xcode_${{ matrix.xcode }}.app/Contents/Developer
-      - name: Upload
-        uses: actions/upload-release-asset@v1.0.1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          upload_url: ${{ github.event.release.upload_url }}
-          asset_path: ./swift-doc
-          asset_name: swift-doc-${{ github.event.release.tag_name }}.catalina.bottle.tar.gz
-          asset_content_type: application/gzip
-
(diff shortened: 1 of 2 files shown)


mattt:
The original `publish` flow also gets the bottling process totally wrong. I'd like to fix that in this PR.

MaxDesiatov (approved):
This is great! I didn't know that preparing `brew bottle` was so easy, I previously had an impression that's too hard to maintain. Now there's a great example for a workflow setup 🙂

mattt:
@MaxDesiatov I know, right? Homebrew is centralized by default, and any documentation I found about the process was specific to their [Brew Test Bot](https://docs.brew.sh/Brew-Test-Bot) service. Back when it launched in 2013, virtualized macOS instances were few and far between, so I can see why that was necessary. But now in 2020 we have GitHub Actions and other infrastructure to stand that up ourselves. I'm not entirely sure when Homebrew started supporting this workflow, but I think this pattern could really start to catch on.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,80 @@
+name: Publish
+
+on:
+  release:
+    types: [created, published]
```
MaxDesiatov, line 5: Do you know what's the difference between these two? It doesn't seem to be documented in [the trigger events doc](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release).

mattt: I'm not 100% sure, but GitHub's UI copy gives us some clue. On the releases page, the button says "Draft a new release". On that page, there are two buttons, "Publish release" or "Save as draft". The former likely corresponds to a `published` event, and the latter may trigger a `created` step.

Here, we're creating a release through the `github/create-release` action, whose name implies a `created` event, but I vaguely recall triggering `published` instead. It's really too bad that there's no easy way to do a dry run of this.

PR #195, opened 2020-10-01
Fix public extensions exposing nested code of all access levels

### LINKED ISSUE #194: Internal variables from namespace extensions are exposed in the generated documentation
When running swift-doc on code block displayed below, the internal variables from namespace extensions get exposed in the generated documentation. There will not be documentation generated for the internal extensions though.

Reproducable with 1.0.0.beta.5

```
public class RootController {
    internal init() {}

    public static func build() -> RootController {
        return RootController()
    }
}


public extension RootController {
    class ControllerExtension {
        public init() {}
        
        public var public_properties: ExtendedProperties = ExtendedProperties()

        internal var internal_properties: InternalProperties = InternalProperties()
    }
}


public extension RootController.ControllerExtension {
    struct ExtendedProperties {
        public init() {}
        public var public_prop: Int = 1
    }
}


internal extension RootController.ControllerExtension {
    struct InternalProperties {
        internal init() {}
        internal var internal_prop: String = "FOO"
    }
}
```

If you run swift-doc on this code, it will result in `RootController_ControllerExtension.md` exposing the variable `internal_properties`, but there will not be a documentation file `RootController_ControllerExtension_InternalProperties.md`. There will be one for `RootController_ControllerExtension_ExtendedProperties` which is public.

I can see that there are some PRs / Issues raised around more granular control over the documentation based on given access level or options to exclude certain symbols from the documentation. However, in my view this is related but different; if the generation is supposed to only process open/public interfaces by default, this behaviour is not met with the included example.

Tunous:
This pull request fixes issue where extensions marked as public would result in all nested members (no matter their access level) being added to the generated documentation. 

I've included 2 tests with my fix, one based on linked issue and one with all simple cases that I've found. I'm not sure whether I've placed them in a correct location though.

Closes #194


Sources/SwiftDoc/Symbol.swift
```diff
@@ -43,7 +43,11 @@ public final class Symbol {
 
         if let `extension` = `extension`,
             `extension`.modifiers.contains(where: { $0.name == "public" }) {
-            return true
+
+            let isNotLimitedAccessModifier = { (modifier: Modifier) in
+                modifier.name != "internal" && modifier.name != "fileprivate" && modifier.name != "private"
+            }
+            return api.modifiers.allSatisfy(isNotLimitedAccessModifier)
```
MaxDesiatov, line 50: **nitpick:** as this closure is not reused anywhere else, why not inline it? The explicit argument type signature wouldn't be needed then:

```suggestion
            return api.modifiers.allSatisfy { modifier in
                modifier.name != "internal" && modifier.name != "fileprivate" && modifier.name != "private"
            }
```

MaxDesiatov (commented):
Overall LGTM, just a single nitpick 🙂

Sources/SwiftDoc/Symbol.swift
Tunous: Sure thing. Originally I wanted to extract this out of this place but then figured out that it’s not necessary and it stayed like this.

MaxDesiatov: Thank you!

MaxDesiatov (approved):

mattt (changes requested):
Thanks for your help with this, @Tunous! I was curious to see whether these changes accommodated the possibility of multiple access modifiers, such as in the case of properties with `private(set)`. With b7ee28f2b0eb017ab3b7a59cafaad7740c259378, I added some failing test cases in `testComputedPropertiesWithMultipleAccessModifiersInPublicExtension`. 

For example, I believe all of the following declarations should be included in the documented interface, because they have public getters (then again, I could be wrong — Swift has some seriously complex rules around ACL):

```swift
public extension Int {
    internal(set) var a: Int {
        get { 1 }
        set {}
    }
    private(set) var b: Int {
        get { 1 }
        set {}
    }
    public internal(set) var c: Int {
        get { 1 }
        set {}
    }
    public private(set) var d: Int {
        get { 1 }
        set {}
    }
}
```

Tunous:
That is an interesting finding. I completely forgot that you could modify access level of properties like this. Having these cases included in documentation sounds like a logical thing to do. I'll try to investigate this topic a little bit further and include necessary changes later.

Tunous:
@mattt I've pushed a fix and a missing test case. Looking at the grammar for [access modifiers](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#grammar_access-level-modifier) it looks like now we are covering all the possible situations.

mattt (approved):
This looks great. Thanks for your help with this, @Tunous!

Eventually, we'll want to codify the [access-level grammar](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#grammar_access-level-modifier) into an enumeration, rather than doing string comparisons (in fact, this is being investigated in #91).

But for now, this certainly does the job. And with the new unit tests, we can go into any future refactoring confident that things will continue to work as they should.

Issue #198, opened 2020-10-04
`all.css` file doesn't contain actual CSS


MaxDesiatov:
Steps to reproduce (steps for purely macOS builds not available due to #197):

1. Run `docker run --entrypoint /bin/bash -it swiftdoc/swift-doc`
2. In the container shell prompt run `apt-get update && apt-get install git`
3. Run `git clone https://github.com/swiftwasm/JavaScriptKit && cd JavaScriptKit`
4. Run `swift-doc generate --base-url "/JavaScriptKit" --module-name JavaScriptKit --format html Sources`
5. Run `cat .build/documentation/all.css`

Expected result: valid CSS.

Actual result: `  Found` output, [can also be seen deployed to GitHub Pages](https://swiftwasm.github.io/JavaScriptKit/all.css).

mattt:
Sorry for the inconvenience, @MaxDesiatov. Swift Package Manager resources caused a bunch of problems for the beta 5 release. I just reverted #192, and am in the process of cutting a new release. What a mess 😞

MaxDesiatov:
No worries. I wonder if we could add an end-to-end test checking that `all.css` has some valid CSS output?

mattt:
@MaxDesiatov Absolutely. I think a lot more end-to-end testing would help avoid botched releases in the future.

MaxDesiatov:
This is fixed in `master` now, thank you!

PR #199, opened 2020-10-04
Test that `all.css` contains a valid CSS selector

MaxDesiatov:
Verifies that generated `all.css` contains the `:root` string on CI. It doesn't do any actual CSS parsing to verify its full validity, only does a simple `grep` check, but I hope this will be enough to prevent issues like #198 in the future.


mattt:
Thanks for getting started on this, @MaxDesiatov. I agree with your instinct to start with a shell script for this. Given that the primary offering of `swift-doc` is the CLI, do you think that it'd make sense to invest in a formal Swift test target that exercises this and other behavior of the executable?

MaxDesiatov:
Sure, let's do that. I'll update the PR accordingly 👍

MaxDesiatov:
@mattt ready for review 🙂

mattt:
@MaxDesiatov Nice! Taking a look now.

Tests/EndToEndTests/EndToEndTests.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    let process = Process()
+    process.executableURL = productsDirectory.appendingPathComponent("swift-doc")
+    process.arguments = ["generate", "--module-name", "SwiftDoc", "--format", "html", "Sources"]
+
+    let pipe = Pipe()
+    process.standardOutput = pipe
+
+    try process.run()
+    process.waitUntilExit()
```
mattt, line 19:
```suggestion
    let process = Process()
    process.executableURL = productsDirectory.appendingPathComponent("swift-doc")
    process.arguments = ["generate", "--module-name", "SwiftDoc", "--format", "html", "Sources"]

    let pipe = Pipe()
    process.standardOutput = pipe

    if #available(OSX 10.13, *) {
      try process.run()
    } else {
      process.launch()
    }
    process.waitUntilExit()
```

mattt: As far as I can tell, the only macOS 10.13 API here is `process.run`, correct? If so, could we tailor the availability check to conditionally use `process.launch` instead where `run` isn't available?

Tests/EndToEndTests/EndToEndTests.swift
```diff
@@ -0,0 +1,50 @@
...
+      .deletingLastPathComponent()
+      .deletingLastPathComponent()
+      .appendingPathComponent("documentation")
+      .appendingPathComponent("all.css")
+    let cssData = try Data(contentsOf: cssPath)
+    guard let css = String(data: cssData, encoding: .utf8) else {
+      return XCTFail("Failed to decode a UTF-8 string from `cssData` in \(#function)")
+    }
+
+    XCTAssertTrue(css.starts(with: ":root"))
```
mattt, line 36:
```suggestion
    XCTAssertTrue(css.contains(":root"))
```

mattt: I could see this assumption breaking unexpectedly, for example if we were to add a header comment or the CSS preprocessor tool decided to reorder or remove the rules. `contains` would be a suboptimal choice if we were worried about performance, but I think this is fine here.

mattt (changes requested):
A few suggested changes for your consideration. Would it be alright if I iterate on this some more this morning, and add some more test cases (e.g. to handle CSS and HTML formats)?

MaxDesiatov:
Sure, please do!

Tests/EndToEndTests/EndToEndTests.swift
```diff
@@ -4,18 +4,18 @@ import XCTest
...
-
     let process = Process()
     process.executableURL = productsDirectory.appendingPathComponent("swift-doc")
     process.arguments = ["generate", "--module-name", "SwiftDoc", "--format", "html", "Sources"]
 
     let pipe = Pipe()
     process.standardOutput = pipe
 
-    try process.run()
+    if #available(OSX 10.13, *) {
```
mattt, line 13: Looks like I missed another macOS 10.13 change: https://github.com/SwiftDocOrg/swift-doc/runs/1215042760?check_suite_focus=true#step:5:67

Making that change now.

Changelog.md
```diff
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added an end-to-end test for validity of generated `all.css` file.
-  #199 by @MaxDesiatov.
+- Added end-to-end tests for generate command-line interface.
```
MaxDesiatov, line 12:
```suggestion
- Added end-to-end tests for `generate` command-line interface.
```

mattt (approved):

Changelog.md
mattt: On second thought, better to remove mention of `generate` entirely, since we're now testing all subcommands.

mattt:
@MaxDesiatov Any last changes before this gets merged?

MaxDesiatov:
LGTM, thank you for finishing this up!

MaxDesiatov:
Can I merge this now? Do you prefer merge commits or linear history by the way?

mattt:
@MaxDesiatov I like to squash and merge PRs. Feel free to do the honors of hitting that green button 😄 (and thanks again for your help with this!)

PR #200, opened 2020-10-06
Specify names for macos and linux jobs

mattt:
Currently, CI workflow results look like this:

Screen Shot 2020-10-06 at 10 51 44

These names are automatically generated based on the job key and the matrix value (Xcode 11.7 and 12 for `macos`; Swift 5.2 and 5.3 for `linux`). This PR uses interpolated matrix values to denote Xcode and Swift versions in the job names for clarity (if that indeed works).


mattt:
Looks like this is working as expected.

Screen Shot 2020-10-06 at 10 54 23

PR #204, opened 2020-10-16
Preserve Formatting of Symbol Declarations

### LINKED ISSUE #201: Avoid horizontal scrolling in formatted code
Currently, formatted code in `swift-doc` requires horizontal scrolling in a lot of places. See attached screenshot from [ComposableArchitecture package docs](https://pointfreeco.github.io/swift-composable-architecture/IfLetStore/).

It would be great if `swift-doc` could reformat the code to some narrow width, at least appropriate for the default font. Something like [SwiftFormat](https://github.com/nicklockwood/swiftformat) could help with this, as it gained line width reformatting capabilities recently.

Screenshot 2020-10-07 at 10 04 33

mattt:
Currently, code declarations for symbols are generated from the `description` of SwiftSemantics objects constructed from syntax nodes. However, this removes any inter-node trivia like newlines that may help break up long lines. 

After considering running the code through a formatter like `swift-format`, I decided that a better approach would be to preserve the existing whitespace, working on the assumption that developers will adopt formatting conventions that look good in their editor.

This PR also tweaks the CSS rules for declarations to better accommodate standard 80 character column widths. Lines up to that width should display without a line break when viewed on desktop browsers at the standard text scale. Longer lines will now wrap according to `pre-wrap` behavior (see https://github.com/SwiftDocOrg/swift-doc/pull/144#issue-456617433), instead of overflowing with a scroll bar.

Resolves #201 

**Before**

Screen Shot 2020-10-16 at 10 12 28

**After**

Screen Shot 2020-10-16 at 10 12 19

Screen Shot 2020-10-16 at 10 13 59


mattt:
One last thing worth mentioning: This change has the incidental benefit of reducing a String -> AST -> HTML conversion step. The current / old way of doing things involved doing a syntactic parse in `SourceFile` to collect symbols, and then doing a separate parse in the HTML post-processing step to do syntax highlighting. With this PR, we now feed the AST nodes from `SourceFile` directly into `SwiftSyntaxHighlighter`, which should make things run a bit faster.

MaxDesiatov (approved):
Seems legit 👍

PR #206, opened 2020-11-09
Use PAT instead of GITHUB_TOKEN to access another repo

kateinoigakukun:
We [carton team](https://github.com/swiftwasm/carton) uses a similar workflow of swift-doc and we found that `NSHipster/update-homebrew-formula-action` needs a GitHub token which has writing permission to tap repository.

[`secrets.GITHUB_TOKEN` doesn't have such permission](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token), so we need to use Personal Access Token instead.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,7 +16,7 @@ jobs:
           tap: SwiftDocOrg/homebrew-formulae
           formula: Formula/swift-doc.rb
         env:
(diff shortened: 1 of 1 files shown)


mattt:
Apologies again for the inconvenience, @kateinoigakukun. I'm hopeful that this is now resolved with https://github.com/NSHipster/update-homebrew-formula-action/pull/6.

PR #221, opened 2021-02-25
Also document default implementations of protocols.

### LINKED ISSUE #213: swift-doc doesn't seem to detect protocol extension methods
We have a library that centers around a single protocol that is extended with methods (similar to how Combine has a `Publisher` protocol and extends it with `map` and other operations), but these methods don't appear to get picked up by swift-doc

Here's documentation for the protocol: https://pointfreeco.github.io/swift-parsing/Parser/

And here's an example extension method: https://github.com/pointfreeco/swift-parsing/blob/6443f30e5766c2319b58412e7c2687e363b54ff1/Sources/Parsing/Parsers/Map.swift#L1-L15

Lukas-Stuehrk:
With this change, we also document default implementations of protocols (and extension methods of protocols). 

This should fix #213

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   #199 by @MaxDesiatov and @mattt.
 - Added `--minimum-access-level` option to `generate` and `coverage` commands.
   #219 by @Lukas-Stuehrk.
+- Added support for documenting default implementations.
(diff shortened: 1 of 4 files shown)


mattt (approved):
Great work, @Lukas-Stuehrk! Sorry for not taking a look at this sooner. I added a changelog entry for this in 879e4e6be66e367457530a246d091ec4a43e0b28, so this should be all clear to merge.

PR #223, opened 2021-03-05
Update README.md to include installation via Mint

ldstreet:
Add instructions to install via Mint https://github.com/yonaskolb/Mint

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,12 @@ $ apt-get update
 $ apt-get install -y libxml2-dev graphviz
 ```
 
+### [Mint](https://github.com/yonaskolb/Mint)
+
+````terminal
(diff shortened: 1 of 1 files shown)


mattt:
Thanks for the PR, @ldstreet. This was previously discussed in #125.

ldstreet:
Oh, didn't see that! Thanks for linking. 

We use a mintfile in our project so just adding a new line there is ideal for us. But that works regardless of the documentation 😀 

In the past we used a Brewfile but one issue that might have been solved by now is that it installed the executables locally which doesn't work well for a build machine that might have many branches with potentially changing versions. Mint gives the ability to run multiple versions of the same tool. 

Thanks for the quick response and your work on this project!

PR #228, opened 2021-03-22
Group operators 

### LINKED ISSUE #123: Group operators separately
Right-now they're just mixed in the same section as normal methods. I think they're special enough to deserve a separate section.

Screenshot 2020-05-21 at 17 17 24

Maybe also the rendering could be improved. A space after the function name and maybe fading out the arguments, as they're not important, or making the operator character bold.

Lukas-Stuehrk:
With this change, operators are separated from methods. Additionally, they are displayed with only their operator.

In the sidebar of the HTML page:
Screenshot 2021-03-22 at 22 30 51

And in the documentation:
Screenshot 2021-03-22 at 22 30 59

This resolves #123.


mattt (commented):
This is off to a great start!

What are you thinking as far as how to distinguish among `prefix` / `infix` / `postfix` operators in the generated UI?

Lukas-Stuehrk:
I'm lacking the imagination how the visuals for this could look like. 

One idea: We could use the type names display the operators with their type names. E.g. we would display an infix operator `==` then like `LeftType == RightType`, a prefix operator `<--` would become `<--Type` and a postfix operator `-->` would become `Type-->`. 

However, this omits the return type of the operation, but the return type might be rather important for custom DSLs with custom operators.

Regarding the implementation then: My current approach simply changed the `name` property which we introduced in an extension on `Function`. This does not feel right for doing this. How do you feel about introducing a `displayedNamed` property?

Also, it's rather easy to distinguish between prefix, infix and postfix operators. The first and the latter always need to have a modifier in the method implementation. But should we add this kind of logic into swift-doc or should we extend SwiftSemantics?

mattt:
Thanks again for taking a stab at this, @Lukas-Stuehrk.

After digging into this a bit more, I realized that I'd originally conflated operators with functions that implement an operator. This became a lot clearer when I tried running `swift-doc` on [Euler](https://github.com/mattt/euler).

To clarify, this is an operator:

```swift
infix operator ≈ : ComparisonPrecedence
```

...and this is a function implementing that operator:

```swift
public func ≈ <T>(lhs: T, rhs: T) -> Bool where T : FloatingPoint {
    return lhs == rhs || lhs.nextDown == rhs || lhs.nextUp == rhs
}
```

I think your change is the correct — functions that implement an operator should be grouped in type pages.

My latest changes build on this by doing the following:

- Update `SourceFile` to collect `OperatorDecl` syntax nodes
- Create a new `OperatorPage`
- Distinguish between infix, prefix, and postfix operators
- Change the home page to order operators by their kind (prefix, etc.)

I also updated the CSS to add a custom style for operators. Here's what all of that looks like:

Operator page:

Screen Shot 2021-04-20 at 03 33 51

Home page:

Screen Shot 2021-04-20 at 03 40 33

PR #229, opened 2021-03-25
Improve file enumeration in generate subcommand

mattt:
This PR makes a few small changes to improve the ergonomics and performance of the `generate` subcommand:

First, we pass `.isDirectoryKey` into the `includingPropertiesForKeys` argument when constructing a file enumerator. I haven't profiled this, but it should be faster for the file manager to pick this up during enumeration rather than fetching it in the loop on-demand.

Second, we pass `.skipsHiddenFiles` and `.skipsPackageDescendants` as enumeration options. Among other things, this prevents cloned dependencies in `.swiftpm` from being pulled in when passing the current directory (`.`) as an input argument.

Third, we tell the file enumerator to skip the top-level `Tests/` directory. In the future, we may replace this with code that parses the `Package.swift` manifest file and identifies the source files directly.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed display of code declarations in HTML.
   #204 by @mattt.
+- Changed the `generate` command to skip hidden files
+  and top-level `Tests` directories.
+  #229 by @mattt.
 
 ## [1.0.0-beta.5] - 2020-09-29
 
--- a/Sources/SwiftDoc/Module.swift
+++ b/Sources/SwiftDoc/Module.swift
@@ -34,7 +34,7 @@ public final class Module {
                        url.lastPathComponent == "Tests",
                        directory.appendingPathComponent("Tests").path == url.path
                     {
-                        directoryEnumerator.skipDescendents()
+                        directoryEnumerator.skipDescendants()
                     }
 
                     continue
--- a/Sources/SwiftDoc/Module.swift
+++ b/Sources/SwiftDoc/Module.swift
@@ -22,7 +22,7 @@ public final class Module {
         let fileManager = FileManager.default
         for path in paths {
             let directory = URL(fileURLWithPath: path)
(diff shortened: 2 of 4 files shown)


Sources/SwiftDoc/Module.swift
```diff
@@ -22,15 +22,27 @@ public final class Module {
...
-                else { continue }
-                sources.append((url, directory))
+                    fileManager.fileExists(atPath: url.path, isDirectory: &isDirectory)
+                else {
+                    // Skip top-level Tests directory
+                    if isDirectory.boolValue == true,
+                       url.lastPathComponent == "Tests",
+                       directory.appendingPathComponent("Tests").path == url.path
+                    {
+                        directoryEnumerator.skipDescendents()
```
Lukas-Stuehrk, line 37: Fun fact. Both `skipDescendents()` and `skipDescendants()` exist. And they should not make a difference.

> This method is identical to skipDescendents() except for the spelling.

But Swift Foundation only implements the `skipDescendants()` method, so we need to use `skipDescendants()`, otherwise it's broken on Linux.

Changelog.md
```diff
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed display of code declarations in HTML.
   #204 by @mattt.
+- Changed the `generate` command to skip hidden files
+  and top-level `Tests` directories.
```
Lukas-Stuehrk, line 29: I don't know how surprising this might be for users. I don't expect anyone to try to build documentation for their tests, but maybe we should mention this in the README?

Lukas-Stuehrk (approved):

Sources/SwiftDoc/Module.swift
mattt: That's... incredible. 🤣 

Thanks for the correction here. I could see myself having spent far too much time trying to track this down.

Changelog.md
mattt: That's a great suggestion. Added with a590cad.

PR #230, opened 2021-03-27
Build documentation for extensions on external types.

Lukas-Stuehrk:
With this change, swift-doc also generates documentation for symbols declared in extensions on external types. This implements #122.

On the homepage, every external type is listed (in a visually very uninspired way) in a new section "Extensions". The following screenshots shows extensions on two different classes of UIKit. One time the external type is referenced with its full name, including the module. One time the external type is referenced only with its class name.
Screenshot 2021-03-27 at 23 21 58

For every external type, a new page is generated and lists the symbols, similar to the already existing pages for types.
Screenshot 2021-03-27 at 23 21 08

Not changed is the behavior for structs and classes which are declared in extensions on external types. Swift-doc already creates documentation for these types.

### DIFF
--- a/Sources/SwiftDoc/Interface.swift
+++ b/Sources/SwiftDoc/Interface.swift
@@ -163,20 +163,33 @@ public final class Interface {
     // MARK: -
 
     public func symbols(named name: String, resolvingTypealiases: Bool) -> [Symbol] {
+        symbolsGroupedByIdentifier.named(name, resolvingTypealiases: resolvingTypealiases)
+    }
+}
+
+fileprivate extension Dictionary where Key == Identifier, Value == [Symbol] {
+    func named(_ name: String, resolvingTypealiases: Bool) -> [Symbol] {
         var pathComponents: [String] = []
         for component in name.split(separator: ".") {
             pathComponents.append("\(component)")
             guard resolvingTypealiases else { continue }
 
-            if let symbols = symbolsGroupedByQualifiedName[pathComponents.joined(separator: ".")],
+            if let symbols = first(where: { $0.key.pathComponents == pathComponents })?.value,
                let symbol = symbols.first(where: { $0.api is Typealias }),
                let `typealias` = symbol.api as? Typealias,
                let initializedType = `typealias`.initializedType
             {
-                pathComponents = (symbol.context.compactMap { ($0 as? Symbol)?.name ?? ($0 as? Extension)?.extendedType }) + [initializedType]
+                let initializedTypePathComponents = initializedType.split(separator: ".")
+                let candidates = keys.filter { $0.matches(initializedTypePathComponents) }
+
+                if let id = candidates.max(by: { $0.pathComponents.count > $1.pathComponents.count }) {
+                    pathComponents = id.pathComponents
+                } else {
+                    return []
+                }
             }
         }
 
-        return symbolsGroupedByQualifiedName[pathComponents.joined(separator: ".")] ?? []
+        return first(where: { $0.key.pathComponents == pathComponents })?.value ?? []
     }
 }
--- a/Sources/SwiftDoc/Interface.swift
+++ b/Sources/SwiftDoc/Interface.swift
@@ -164,7 +164,7 @@ public final class Interface {
 
     public func symbols(named name: String, resolvingTypealiases: Bool) -> [Symbol] {
         var pathComponents: [String] = []
-        for component in name.split(separator: ".", omittingEmptySubsequences: true) {
+        for component in name.split(separator: ".") {
             pathComponents.append("\(component)")
             guard resolvingTypealiases else { continue }
 
--- a/Sources/SwiftDoc/Interface.swift
+++ b/Sources/SwiftDoc/Interface.swift
@@ -10,16 +10,18 @@ public final class Interface {
         self.imports = imports
         self.symbols = symbols
 
-        self.symbolsGroupedByIdentifier = Dictionary(grouping: symbols, by: { $0.id })
-        self.symbolsGroupedByQualifiedName = Dictionary(grouping: symbols, by: { $0.id.description })
-        self.topLevelSymbols = symbols.filter { $0.api is Type || $0.id.pathComponents.isEmpty }
+        let symbolsGroupedByIdentifier = Dictionary(grouping: symbols, by: { $0.id })
+        let symbolsGroupedByQualifiedName = Dictionary(grouping: symbols, by: { $0.id.description })
+
+        self.symbolsGroupedByIdentifier = symbolsGroupedByIdentifier
+        self.symbolsGroupedByQualifiedName = symbolsGroupedByQualifiedName
+        self.topLevelSymbols = symbols.filter { $0.api is Type || $0.id.context.isEmpty }
 
         self.relationships = {
             let extensionsByExtendedType: [String: [Extension]] = Dictionary(grouping: symbols.flatMap { $0.context.compactMap { $0 as? Extension } }, by: { $0.extendedType })
 
             var relationships: Set<Relationship> = []
             for symbol in symbols {
-
                 let lastDeclarationScope = symbol.context.last(where: { $0 is Extension || $0 is Symbol })
                 
                 if let container = lastDeclarationScope as? Symbol {
@@ -40,8 +42,7 @@ public final class Interface {
                 }
 
                 if let `extension` = lastDeclarationScope as? Extension {
-                    if let extended = symbols.first(where: { $0.api is Type &&  $0.id.matches(`extension`.extendedType) }) {
-
+                    for extended in symbolsGroupedByIdentifier.named(`extension`.extendedType, resolvingTypealiases: true) {
                         let predicate: Relationship.Predicate
                         switch extended.api {
                         case is Protocol:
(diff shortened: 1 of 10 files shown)


Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -58,14 +58,17 @@ extension SwiftDoc {
 
         var pages: [String: Page] = [:]
 
+        var declaredTypeNames: [String] = []
```
mattt, line 61: Is there any reason why we're collecting them here, rather than querying `module.interface` for symbols for which `symbol is Type || symbol is TypeAlias`?

Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -75,6 +78,17 @@ extension SwiftDoc {
           }
         }
 
+        // Extensions on external types.
+        var symbolsByExternalType: [String: [Symbol]] = [:]
+        for symbol in module.interface.symbols.filter(symbolFilter) {
+          guard let ext = symbol.context.first as? Extension, symbol.context.count == 1 else { continue }
+          guard !declaredTypeNames.contains(ext.extendedType) else { continue }
```
mattt, line 85: It looks like this approach is generating false positives. I tried running this against our [SwiftMarkup](https://github.com/SwiftDocOrg/SwiftMarkup) project, and it showed `Callout.Delimiter` as an extended type. 
This code appears to be picking up the conformance to `Codable` and `RawValueRepresentable` adopted in an extension in the same file.

I haven't dug into this too deeply, but my guess is that this has to do with a difference between base and qualified symbol names.

mattt (changes requested):
Thanks for taking this on, @Lukas-Stuehrk.

I found an issue with the current behavior when I was testing this locally. See my comments in  `Subcommands/Generate.swift`.

In other news, I added a commit that improves how the extensions are styled. 

Screen Shot 2021-03-29 at 06 06 56

Sources/swift-doc/Subcommands/Generate.swift
Lukas-Stuehrk: There was a reason which I don't remember exactly anymore. I think filtering a second time put quite some additional time in large code bases (e.g. bigger iOS apps and not only small libraries).

However, I also think that this spreads into unrelated logic and makes the code harder to understand. As we don't have too many performance improvements yet, maybe we can not spread the logic, filter a second time and then put some focus on performance in the future.

Lukas-Stuehrk: Okay, forget about all of this.

It's rather easy and definitely the better approach to use all information which we have available on `Interface`, e.g. `symbolsGroupedByQualifiedName`. This will not only remove the spreading of the logic, it will also enable better detection of external symbols.

I'll pick this up later this week.

Sources/swift-doc/Subcommands/Generate.swift
Lukas-Stuehrk: My new approach is only _slightly_ crazy. But it has surprisingly good results for a tool that does not understand semantics.

mattt:
@Lukas-Stuehrk I'm interested to hear what you think of my changes in 89f5e83.

Reading through your implementation of `isExternalSymbol`, I realized that we could generalize this functionality to resolve any qualified name through any typealiases that exist. Currently, we're only using it to determine whether a named symbol is external, in which case the equivalent to `isExternalSymbol(name)` is now `symbols(named: name, resolvingTypealiases: true).isEmpty`. Beyond that, I think this could be key to fixing some of the limitations we've had with our hand-rolled parser.

This has all of your original tests passing, with a few additional ones. How does it look for you?

Sources/SwiftDoc/Interface.swift
```diff
@@ -159,4 +159,24 @@ public final class Interface {
     public func defaultImplementations(of symbol: Symbol) -> [Symbol] {
         return relationshipsByObject[symbol.id]?.filter { $0.predicate == .defaultImplementationOf }.map { $0.subject }.sorted() ?? []
     }
+
+    // MARK: -
+
+    public func symbols(named name: String, resolvingTypealiases: Bool) -> [Symbol] {
+        var pathComponents: [String] = []
+        for component in name.split(separator: ".", omittingEmptySubsequences: true) {
```
Lukas-Stuehrk, line 167: Nitpick: `omittingEmptySubsequences` defaults to true and can be omitted.

Sources/SwiftDoc/Interface.swift
```diff
@@ -159,4 +159,24 @@ public final class Interface {
...
+            pathComponents.append("\(component)")
+            guard resolvingTypealiases else { continue }
```
Lukas-Stuehrk, line 169: 🤔 This is sooooo clever.

Lukas-Stuehrk:
To be transparent: There are still some shortcomings. One example:

```swift
import UIKit 
            
public class SomeClass {
    public struct InnerObject { }
   
    typealias ActuallyExternal = UIView
            
    typealias ActuallyInternal = InnerStruct 
    
    struct InnerStruct {}        
}
            
public class AnotherClass {
    typealias PublicClass = SomeClass            
}
            
```

In this example, the lookup for `AnotherClass.PublicClass.ActuallyInternal` does not work. 

But nevertheless, I think we cover all of the most common cases, especially for libraries. And this is already a huge improvement. And for me it was really surprising how far we can get with a "cheap" and completely non-semantic symbol lookup.

And I fully agree, this could solve some of the existing problems in the symbol resolution and I like your generalization of the problem. But it also requires to think of what would be the expected result for many cases. One example again:

```swift
public class SomeClass {

}

public typealias AnotherClass = SomeClass

public extension AnotherClass {
    func methodOfExtension() {

    }
}
```

In this example, `swift-doc` does not include `methodOfExtension()` anywhere in the generated documentation (and I think fixing this should be outside the scope of this PR). But where should we include it? Simply add it to `SomeClass`? What was the intention of the author to express it like this? How do we reflect the intention in the generated documentation or do we even need to?

Sources/SwiftDoc/Interface.swift
mattt:
```suggestion
        for component in name.split(separator: ".") {
```

mattt:
> To be transparent: There are still some shortcomings. One example:
> 
> ```swift
> import UIKit 
>             
> public class SomeClass {
>     public struct InnerObject { }
>    
>     typealias ActuallyExternal = UIView
>             
>     typealias ActuallyInternal = InnerStruct 
>     
>     struct InnerStruct {}        
> }
>             
> public class AnotherClass {
>     typealias PublicClass = SomeClass            
> }
>             
> ```
> 
> In this example, the lookup for `AnotherClass.PublicClass.ActuallyInternal` does not work.

Thanks for sharing that failing test case. I managed to add support for this in bac46a1. The basic idea is that, when we encounter a `typealias` in a key path, we resolve the alias to a matching symbol and then continue with the remaining components from that new reference point.

For example, consider the following:

```swift
enum A {
  typealias B = X
}

enum X { enum Y { enum Z { } } }
```

If we want to resolve `"A.B.Y.Z"`, we walk to `A`, then `A.B` and encounter a `typealias`. The initialized type name `"X"` is resolved to the top-level `X`, so from here, the operation is equivalent to resolving `"X.Y.Z"`. 

> But nevertheless, I think we cover all of the most common cases, especially for libraries. And this is already a huge improvement. And for me it was really surprising how far we can get with a "cheap" and completely non-semantic symbol lookup.

💯. I'm sure this could be more robust, but I'm happy with this for now if it solves our current problem.

> And I fully agree, this could solve some of the existing problems in the symbol resolution and I like your generalization of the problem. But it also requires to think of what would be the expected result for many cases. One example again:
> 
> ```swift
> public class SomeClass {
> 
> }
> 
> public typealias AnotherClass = SomeClass
> 
> public extension AnotherClass {
>     func methodOfExtension() {
> 
>     }
> }
> ```
> 
> In this example, `swift-doc` does not include `methodOfExtension()` anywhere in the generated documentation (and I think fixing this should be outside the scope of this PR). But where should we include it? Simply add it to `SomeClass`? What was the intention of the author to express it like this? How do we reflect the intention in the generated documentation or do we even need to?

These are all great questions. I'll open a new issue to track this for a future release.

Lukas-Stuehrk:
I started to write some tests for your changes and found yet another and arguably extreme  test case which fails 🙈:

```swift
    public func testMembersOfTypealiasedSymbols() throws {
        let source = #"""
        public class SomeClass {
            public typealias IndirectSelfAlias = OtherClass
                
            public func someMethod() { }
        }
                
        public typealias OtherClass = SomeClass
                
        public extension OtherClass {
            func someExtensionMethod() { }
        }
                
        public typealias IndirectSomeClass = OtherClass.IndirectSelfAlias
                
        public extension IndirectSomeClass {
            func anotherExtensionMethod() { }
        }
        """#


        let url = try temporaryFile(contents: source)
        let sourceFile = try SourceFile(file: url, relativeTo: url.deletingLastPathComponent())
        let module = Module(name: "Module", sourceFiles: [sourceFile])

        XCTAssertEqual(module.interface.symbols.count, 7)

        let someClass = module.interface.symbols[0]
        XCTAssertEqual(someClass.name, "SomeClass")
        let members = module.interface.members(of: someClass)
        XCTAssertEqual(4, members.count)
        XCTAssertEqual(members[0].name, "IndirectSelfAlias")
        XCTAssertEqual(members[1].name, "someMethod()")
        XCTAssertEqual(members[2].name, "someExtensionMethod()")
        XCTAssertEqual(members[3].name, "anotherExtensionMethod()")
    }
```

But maybe we should call it a day and consider this a future problem. Then we could go with passing tests like:
```swift
    public func testMembersOfTypealiasedSymbols() throws {
        let source = #"""
        public class SomeClass {
            public func someMethod() { }
        }
                
        public typealias OtherClass = SomeClass
                
        public extension OtherClass {
            func someExtensionMethod() { }
        }
        """#


        let url = try temporaryFile(contents: source)
        let sourceFile = try SourceFile(file: url, relativeTo: url.deletingLastPathComponent())
        let module = Module(name: "Module", sourceFiles: [sourceFile])

        XCTAssertEqual(module.interface.symbols.count, 4)

        let someClass = module.interface.symbols[0]
        XCTAssertEqual(someClass.name, "SomeClass")
        let members = module.interface.members(of: someClass)
        XCTAssertEqual(2, members.count)
        XCTAssertEqual(members[0].name, "someMethod()")
        XCTAssertEqual(members[1].name, "someExtensionMethod()")
    }
```

mattt:
@Lukas-Stuehrk 🙈 

Agreed. Let's lock in our gains and ship this thing. Do you think this is ready to go in today's beta? Or would you like some extra time for it to bake?

Lukas-Stuehrk:
I think this is ready! And it's a nice improvement. I pushed the passing tests.

mattt (approved):

PR #233, opened 2021-03-28
Add a workflow to validate the generated assets on CI.

Lukas-Stuehrk:
It's rather easy to accidentally commit the wrong version of the generated CSS assets. Or to not commit the generated version and only the changed source. And it's not easy to validate if the generated CSS of a pull request correlates to the sources of the pull request.

This change introduces a new pipeline for CI. This pipeline builds the CSS from the sources and validates that the committed version of the generated CSS is identical to building it from source.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,12 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 14.x
+      - uses: actions/cache@v2
+        with:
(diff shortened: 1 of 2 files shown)


mattt (approved):
This is a great addition to our CI workflow. Thanks for taking the time to implement this, @Lukas-Stuehrk.

I added a step to cache NPM dependencies, which appears to be working as expected. Please give that a quick look and then feel free to merge.

Issue #119, opened 2020-05-21
Add "View Source" link


sindresorhus:
It would be nice if each type had a "View Source" link (could be on hover or in a `…` menu) that linked to its source (with the correct SHA reference) on GitHub. That way, users could quickly jump into the source 

As with https://github.com/SwiftDocOrg/swift-doc/issues/117, this would need some kind of `githubRepoURL` config.

PR #237, opened 2021-03-29
Add sourceRange property to Symbol

mattt:
Currently, `Symbol` defines a `sourceLocation` property that returns the start location of the symbol's declaration. As a step towards supporting #119, this PR replaces that with a new `sourceRange` property that comprises the start and end location.


Sources/SwiftDoc/Symbol.swift
```diff
@@ -225,7 +228,7 @@ extension Symbol: Codable {
     private enum CodingKeys: String, CodingKey {
         case declaration
         case documentation
-        case sourceLocation
+        case sourceRange
```
Lukas-Stuehrk, line 231: 🤔 This is also a topic for the "Do we really offer SwiftDoc as a standalone library?" discussion. If yes, we would need to document changes of the serialization of public objects.

Lukas-Stuehrk (approved):
Nice and very useful change, @mattt. (I needed exactly this functionality in a small tool that uses the `SwiftDoc` library and [implemented it not as elegant in a branch of my fork](https://github.com/Lukas-Stuehrk/swift-doc/commit/a0087235ba5620713ae2f657722d19a9250782f3)).

Sources/SwiftDoc/Symbol.swift
mattt: Yeah, that's a fair point. (Although, to be honest, I don't recall why `Symbol` is `Codable` in the first place...)

I added changelog entries with 11b21ed. Anything else to be done here?

Lukas-Stuehrk: Early preparations for caching or for getting symbols from different sources than source files?

For the long-running discussion outside the scope of this pull request: I think removing `Codable` conformance is a reasonable option. It's not needed for swift-doc and adds maintenance effort. Users of the library could implement it themselves.

For this pull request: Looks good to me, but there's now a conflict because of the changelog changes.

mattt: > Early preparations for caching or for getting symbols from different sources than source files?

Yeah, that was probably it. Let's leave things as they are for now. We can revisit all of that when we get to vending serialized representations (such as JSON or RDF).

PR #238, opened 2021-03-29
Update is-svg dependency to >=4.2.2

mattt:


Lukas-Stuehrk (approved):

mattt:
Ah, looks like this triggered the new `Validate Assets` CI job after merging to master. Rebuilt assets with 35b7a82.

Issue #234, opened 2021-03-29
Adding possibility to ignore certain folders


dtamiazzo:
Hi, 
I don't know if it's a feature that's already implemented, but I'd really like to be able to specify certain folders to ignore when creating the documentation. 
Specifically, I need to avoid creating documentations for the Pods I imported in my app.
It would be nice to be able to have a sort of gitignore equivalent, a list of folders to ignore in  an external file.

Thank you

mattt:
@dtamiazzo The approach so far has been to pass subdirectory paths as inputs rather than the root directory of your project. For example, running `swift-doc generate ./MyApp/Sources` instead of `./MyApp` would ignore any source files in `./MyApp/Pods`.

That said, I think most people would like `Pods`, `Tests`, and other common directories to be similarly ignored. To that end, I opened #241. Let me know what you think!

mattt:
Fixed by #241.

PR #241, opened 2021-03-30
Ignore additional top-level directories

mattt:
Related to #234

Building upon #229, which ignores hidden directories `./Tests/`, this PR also ignores the following top-level directories:

- `node_modules`
- `Packages`
- `Pods`
- `Resources`

### DIFF
--- a/Sources/SwiftDoc/Module.swift
+++ b/Sources/SwiftDoc/Module.swift
@@ -22,37 +22,31 @@ public final class Module {
         let fileManager = FileManager.default
         for path in paths {
             let directory = URL(fileURLWithPath: path)
-            guard let directoryEnumerator = fileManager.enumerator(at: directory, includingPropertiesForKeys: [.isDirectoryKey], options: [.skipsHiddenFiles, .skipsPackageDescendants]) else { continue }
-            for case let url as URL in directoryEnumerator {
-                var isDirectory: ObjCBool = false
-                guard url.pathExtension == "swift",
-                    fileManager.isReadableFile(atPath: url.path),
-                    fileManager.fileExists(atPath: url.path, isDirectory: &isDirectory)
-                else {
-                    if isDirectory.boolValue == true {
-                        let ignoredTopLevelDirectories: [String] = [
-                            "node_modules",
-                            "Packages",
-                            "Pods",
-                            "Resources",
-                            "Tests"
-                        ]
+            guard let directoryEnumerator = fileManager.enumerator(at: directory,
+                                                                   includingPropertiesForKeys: [.isDirectoryKey],
+                                                                   options: [.skipsHiddenFiles, .skipsPackageDescendants])
+            else { continue }
 
-                        for topLevelDirectory in ignoredTopLevelDirectories {
-                            if url.lastPathComponent == topLevelDirectory,
-                               directory.appendingPathComponent(topLevelDirectory).path == url.path
-                            {
-                                directoryEnumerator.skipDescendants()
-                            }
-                        }
+            let ignoredDirectories: Set<URL> = Set([
+                "node_modules",
+                "Packages",
+                "Pods",
+                "Resources",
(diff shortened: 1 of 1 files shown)


Sources/SwiftDoc/Module.swift
```diff
@@ -29,12 +29,22 @@ public final class Module {
                     fileManager.isReadableFile(atPath: url.path),
                     fileManager.fileExists(atPath: url.path, isDirectory: &isDirectory)
                 else {
-                    // Skip top-level Tests directory
-                    if isDirectory.boolValue == true,
-                       url.lastPathComponent == "Tests",
-                       directory.appendingPathComponent("Tests").path == url.path
-                    {
-                        directoryEnumerator.skipDescendants()
+                    if isDirectory.boolValue == true {
```
Lukas-Stuehrk, line 32: I'm afraid this won't work. There is a logic bug, which was introduced in #229 and I overlooked it in the code review 🙈:

In line 28: `url.pathExtension == "swift"` is false, because of this all following conditions in the guard statement never get executed. Because of this, `isDirectory` is never updated and is always false, so the directories are not skipped.

To be honest, I find the logic for collection all files very hard to understand by now. Maybe we could make it less complicated and more verbose and go with something like this:
```swift
        for path in paths {
            let directory = URL(fileURLWithPath: path)
            guard let directoryEnumerator = fileManager.enumerator(at: directory, includingPropertiesForKeys: [.isDirectoryKey], options: [.skipsHiddenFiles, .skipsPackageDescendants]) else { continue }
            let ignoredDirectories = Set([
               directory.appendingPathComponent("node_modules"),
               directory.appendingPathComponent("Packages"),
               directory.appendingPathComponent("Pods"),
               directory.appendingPathComponent("Resources"),
               directory.appendingPathComponent("Tests"),
            ])
            for case let url as URL in directoryEnumerator {
                guard url.pathExtension == "swift", fileManager.isReadableFile(atPath: url.path)
                else {
                    if ignoredDirectories.contains(url) {
                        directoryEnumerator.skipDescendants()
                    }

                    continue
                }

                sources.append((url, directory))
            }
        }
```
(We could even

Lukas-Stuehrk (changes requested):
👍 Nice change, I like the idea that the default behavior and configuration for swift-doc does the "right" thing for most projects. Especially since it's still possible to include those directories if needed by providing their full path additionally in the list of paths.

There's a little logic bug, which needs to be fixed, otherwise the directories won't be excluded.

Sources/SwiftDoc/Module.swift
mattt: Thanks for catching that. I like your approach, and followed that with dcdfe9716cc22a3373a19c74f0f224195f5d8bbc.

Sources/SwiftDoc/Module.swift
```diff
@@ -22,27 +22,31 @@ public final class Module {
...
+                                                                   options: [.skipsHiddenFiles, .skipsPackageDescendants])
+            else { continue }
+
+            let ignoredDirectories: Set<URL> = Set([
+                "node_modules",
+                "Packages",
+                "Pods",
+                "Resources",
+                "Tests"
+            ].map { directory.appendingPathComponent($0) })
```
Lukas-Stuehrk, line 36: It's also an option to use`.map(directory.appendingPathComponent))`

Lukas-Stuehrk (approved):

Sources/SwiftDoc/Module.swift
mattt: Indeed! I considered that, but decided that cut against "making things clearer". Then again, are anonymous closure arguments (`$0`) _really_ better? 🙃

mattt:
@Lukas-Stuehrk Seriously, thanks again for catching my mistake here. I'm excited to get this merged and cut a new beta soon.

PR #242, opened 2021-04-03
Display a warning if an input directory does not exist.

Lukas-Stuehrk:
A small improvement, so users can troubleshoot problems easier. 

Before, when you accidentally misspelled the input path or provided for any other reason a path to directory which does not exist, swift-doc only displayed a warning that no symbols where found. This could lead new users into a very wrong direction.

With this change, there's an additional warning that the given directory does not exist. If an input path exists but is not a directory, a helpful warning is displayed as well, as swift-doc only works with paths to directories and not direct paths to Swift source files.

### DIFF
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -47,6 +47,15 @@ extension SwiftDoc {
     var options: Options
 
     func run() throws {
+      for directory in options.inputs {
+        var isDirectory: ObjCBool = false
+        if !FileManager.default.fileExists(atPath: directory, isDirectory: &isDirectory) {
+          logger.warning("Input directory \(directory) does not exist.")
+        } else if !isDirectory.boolValue {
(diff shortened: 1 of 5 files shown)


mattt (approved):
This is an excellent and thoughtful change. Great work, @Lukas-Stuehrk! Merging this in now.

Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -47,6 +47,15 @@ extension SwiftDoc {
     var options: Options
 
     func run() throws {
+      for directory in options.inputs {
+        var isDirectory: ObjCBool = false
+        if !FileManager.default.fileExists(atPath: directory, isDirectory: &isDirectory) {
+          logger.warning("Input directory \(directory) does not exist.")
```
mattt, line 53:
```suggestion
          logger.warning("Input path \(directory) does not exist.")
```

Small change to make message consistent with how we call input arguments "paths"

Issue #202, opened 2020-10-07
Add end-to-end continuous integration tests for Homebrew distribution


mattt:
As discussed in #196 and #197, we should make sure the `swift-doc` executable installed by Homebrew continues to work. Here's what I'm thinking:

- Create a new GitHub Actions workflow that runs on pushes to master as well as on a cron schedule (once per day, perhaps)
- In that action, use the test bundle from #199 to work on `swift-doc` installed by Homebrew instead of the one built from source (this could be configured via a command argument or environment variable)
- Set up the workflow to test `brew install swift-doc` both building from source and downloading from a bottle, if available 
- Add corresponding CI tests in @SwiftDocOrg/homebrew-formulae

mattt:
Fixed by #243.

PR #243, opened 2021-04-04
Add end-to-end continuous integration tests for Homebrew distribution

Lukas-Stuehrk:
This changes the end-to-end tests so it's possible to use the tests with the `swift-doc` version installed via homebrew.

It also adds a new GitHub actions workflow to run the end-to-tests with the homebrew version of swift-doc, both from bottle and from source.

This implements #202.

Unfortunately, the integration tests will fail. I can reproduce this also on my machine when I use the swift-doc version from homebrew:

```bash
lukas@grumpycat ~/src/swift-doc $ which swift-doc                                                                           
/Users/lukas/homebrew/bin/swift-doc
lukas@grumpycat ~/src/swift-doc $ swift-doc generate --module-name SwiftDoc --format commonmark --output /tmp/docs .
[1]    47753 illegal hardware instruction  swift-doc generate --module-name SwiftDoc --format commonmark --output  .
```

The crashing line is https://github.com/SwiftDocOrg/swift-doc/blob/f441648bcb8a6b07a3724bc1af2488e8a6e6c184/Sources/SwiftDoc/SourceFile.swift#L141-L141

### DIFF
--- a/.github/workflows/homebrew_ci.yml
+++ b/.github/workflows/homebrew_ci.yml
@@ -1,4 +1,4 @@
-name: homebrew CI
+name: Homebrew CI
 
 on:
   push:
@@ -9,7 +9,7 @@ on:
 jobs:
   homebrew-bottle:
     runs-on: macos-10.15
-    name: homebrew from bottle
+    name: Test Homebrew installation from bottle
 
     steps:
       - name: Checkout
@@ -29,7 +29,7 @@ jobs:
   homebrew-sources:
     runs-on: macos-11.0
 
-    name: homebrew from source
+    name: Test Homebrew installation from sources
     steps:
       - name: Checkout
         uses: actions/checkout@v1
--- a/.github/workflows/homebrew_ci.yml
+++ b/.github/workflows/homebrew_ci.yml
@@ -1,0 +1,45 @@
+name: homebrew CI
+
+on:
+  push:
+    branches: [master]
+  schedule:
+    - cron: 0 0 * * *
+
+jobs:
+  homebrew-bottle:
+    runs-on: macos-10.15
+    name: homebrew from bottle
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - uses: actions/cache@v2
+        with:
+          path: .build
+          key: ${{ runner.os }}-spm-xcode-${{ matrix.xcode }}-${{ hashFiles('**/Package.resolved') }}
+          restore-keys: |
+            ${{ runner.os }}-spm-xcode-${{ matrix.xcode }}-
+      - name: Install swift-doc
+        run: brew install swiftdocorg/formulae/swift-doc --force-bottle
+      - name: Run Tests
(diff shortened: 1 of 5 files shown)


.github/workflows/homebrew_ci.yml
```diff
@@ -0,0 +1,45 @@
...
+on:
+  push:
+    branches: [master]
+  schedule:
+    - cron: 0 0 * * *
+
+jobs:
+  homebrew-bottle:
+    runs-on: macos-10.15
+    name: homebrew from bottle
```
mattt, line 12:
```suggestion
    name: Test Homebrew installation from bottle
```

.github/workflows/homebrew_ci.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Install swift-doc
+        run: brew install swiftdocorg/formulae/swift-doc --force-bottle
+      - name: Run Tests
+        run: |
+          swift test --filter EndToEndTests. -Xswiftc -DUSE_HOMEBREW
+
+  homebrew-sources:
+    runs-on: macos-11.0
+
+    name: homebrew from source
```
mattt, line 32:
```suggestion
    name: Test Homebrew installation from sources
```

.github/workflows/homebrew_ci.yml
```diff
@@ -0,0 +1,45 @@
+name: homebrew CI
```
mattt, line 1:
```suggestion
name: Homebrew CI
```

mattt:
> Unfortunately, the integration tests will fail. I can reproduce this also on my machine when I use the swift-doc version from homebrew:
> 
> ```shell
> lukas@grumpycat ~/src/swift-doc $ which swift-doc                                                                           
> /Users/lukas/homebrew/bin/swift-doc
> lukas@grumpycat ~/src/swift-doc $ swift-doc generate --module-name SwiftDoc --format commonmark --output /tmp/docs .
> [1]    47753 illegal hardware instruction  swift-doc generate --module-name SwiftDoc --format commonmark --output  .
> ```
> 
> The crashing line is
> 
> https://github.com/SwiftDocOrg/swift-doc/blob/f441648bcb8a6b07a3724bc1af2488e8a6e6c184/Sources/SwiftDoc/SourceFile.swift#L141-L141

Huh, that's surprising. I thought Homebrew built and installed in release mode, so assertions would be disabled. Let me double-check what we're doing in our formula. At the very least, I can remove that assertion — that's something we could and probably should be validating in a unit test instead.

mattt:
> Unfortunately, the integration tests will fail. I can reproduce this also on my machine when I use the swift-doc version from homebrew...

Oh wait, I recognize that crash! That's what happens when you try to run `swift-doc` on `SwiftSyntax` (or one of the other dependencies). Fortunately, this is fixed by #229. We can hold off on merging this until that gets into an official release.

mattt (approved):

Issue #73, opened 2020-04-10
Write a "Getting Started" guide


mattt:
Expanding on what's currently in the README, we should provide a step-by-step tutorial to help onboard new users.

PR #247, opened 2021-04-17
Add a user guide

Ehlster:
This is a discussion starter. It's a first try to explore the structure and contents of a user guide. I _think_ it implements #73.  It uses a lot of functionality of the upcoming `1.0.0-beta.6` release, so we should not merge it before this version is released.

Please feel free to change, extend and rework this. Also, in its current form it's rather boring 🇩🇪 and we might want to make it more exciting.

Also: English is not my first language, so it could benefit from improvements of a native speaker with good writing skills.


davegomez:
I'm looking for something like this.

My doubt is how should I document a return value when is optional?

Should I just mention it in the comment and the generator will take the Optional Type automatically setting the difference somewhere?

PR #255, opened 2021-04-24
git: ignore vim swap files

compnerd:
For those developers who use vim, the traditional behaviour is to create
swap files next to the file itself.  These files should not be committed
to the repository, add them to the ignore list.


mattt (approved):

compnerd:
Created #272 to deal with the signing.

PR #257, opened 2021-04-25
swift-doc: load FoundationNetworking on non-Darwin platforms

compnerd:
Invert the condition to not link against Foundation networking only on
non-Darwin platforms.  Every other platform uses
swift-corelibs-foundation which provides a FoundationNetworking which
must be linked against in order to have the networking functionality
work.  This repairs the fetch of the CSS for Windows hosts.

### DIFF
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -5,7 +5,7 @@ import SwiftMarkup
 import SwiftSemantics
 import struct SwiftSemantics.Protocol
 
-#if os(Linux)
+#if !(os(iOS) || os(macOS) || os(tvOS) || os(watchOS))
 import FoundationNetworking
 #endif
 
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -5,7 +5,7 @@ import SwiftMarkup
 import SwiftSemantics
 import struct SwiftSemantics.Protocol
 
-#if !(os(iOS) || os(macOS) || os(tvOS) || os(watchOS))
+#if canImport(FoundationNetworking)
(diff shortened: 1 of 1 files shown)


Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -5,7 +5,7 @@ import SwiftMarkup
 import SwiftSemantics
 import struct SwiftSemantics.Protocol
 
-#if os(Linux)
+#if !(os(iOS) || os(macOS) || os(tvOS) || os(watchOS))
```
mattt, line 8: According to [this doc in swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation/blob/main/Docs/ReleaseNotes_Swift5.md#dependency-management), the preferred way to check would be

```suggestion
#if canImport(FoundationNetworking)
```

Would that work for Windows?

compnerd: Yes, that should work for Windows. I didn't think of `canImport(FoundationNetworking)`.

MaxDesiatov (approved):
LGTM 👍

compnerd:
Hmm, what is with the signed commit requirement on this repository?  I don't think that I have a setup where configuring that is convenient.  I suppose that I could do this through the web editor :/

MaxDesiatov:
@mattt I would be happy to merge, but the signing requirement prevents me from doing it and I can't override it. I think if you'd like all commits in `master` to be signed, enabling "Squash and Merge" mode as the default should be enough. PRs merged that way are automatically signed by GitHub.

compnerd:
@MaxDesiatov I'm going to close this as I've opened https://github.com/SwiftDocOrg/swift-doc/pull/265 via the web editor which should allow me to merge.  Would you mind approving that?

PR #258, opened 2021-04-25
perform some path sanitization over the generated names

compnerd:
The current scheme for the file generation uses the function label,
which uses ':' as a parameter separator.  However, this character is not
valid on all file systems.  This results in partial content generation
on Windows.  Add a sanitization step which replaces "invalid" characters
with `_` instead.  This does not impact the rendering, only the
filenames and the generated URL references between the pages.  With
this, we can generate and link function documentation on Windows.

### DIFF
--- a/Sources/SwiftDoc/Helpers.swift
+++ b/Sources/SwiftDoc/Helpers.swift
@@ -23,10 +23,11 @@ public func path(for identifier: CustomStringConvertible, with baseURL: String) 
     return string
 }
 
+private let reservedCharacters: CharacterSet = [
+    // Windows Reserved Characters
+    "<", ">", ":", "\"", "/", "\\", "|", "?", "*",
+]
+
 public func path(for identifier: String) -> String {
-    let kReservedCharacters: CharacterSet = [
-      // Windows Reserved Characters
-      "<", ">", ":", "\"", "/", "\\", "|", "?", "*",
-    ]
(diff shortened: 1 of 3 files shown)


Sources/SwiftDoc/Helpers.swift
```diff
@@ -14,10 +14,19 @@ public func path(for symbol: Symbol, with baseURL: String) -> String {
...
     return string
 }
+
+public func path(for identifier: String) -> String {
+    let kReservedCharacters: CharacterSet = [
+      // Windows Reserved Characters
+      "<", ">", ":", "\"", "/", "\\", "|", "?", "*",
+    ]
+    return identifier.components(separatedBy: kReservedCharacters).joined(separator: "_")
+}
```
mattt, line 32:
```suggestion

private let reservedCharacters: CharacterSet = [
    // Windows Reserved Characters
    "<", ">", ":", "\"", "/", "\\", "|", "?", "*",
]

public func path(for identifier: String) -> String {
    return identifier.components(separatedBy: reservedCharacters).joined(separator: "_")
}
```

mattt (approved):

mattt:
@compnerd Thanks for your patience on this. Once CI comes back ✅ , I'll merge this in.

compnerd:
@mattt - do I need to redo the patch or can you merge this (for the signing)?

mattt:
@compnerd Sorry for the delay. This is now merged in.

PR #260, opened 2021-04-25
Fix toplevel symbols

Lukas-Stuehrk:
Looks like [my assessment was wrong](https://github.com/SwiftDocOrg/swift-doc/pull/230#issuecomment-825810114) and we introduced a regression in #230. 

There's a logic error when checking which symbols are considered top-level symbols. The path of a symbol's identifier always contains the symbol's own name. Because of this, `swift-doc` version `1.0.0-beta.6` is not including all top level symbols which are not structs or classes.

This change fixes that regression and adds some tests.

I think we should add some regression tests at some point to check which symbols are included in a generated documentation.

### DIFF
--- a/Sources/SwiftDoc/Interface.swift
+++ b/Sources/SwiftDoc/Interface.swift
@@ -24,7 +24,7 @@ public final class Interface {
                 return false
             }
 
-            return symbol.id.pathComponents.isEmpty
+            return symbol.id.pathComponents.count == 1
         }
 
         self.relationships = {
--- a/Sources/SwiftDoc/Interface.swift
+++ b/Sources/SwiftDoc/Interface.swift
@@ -24,7 +24,7 @@ public final class Interface {
                 return false
             }
 
-            return symbol.id.pathComponents.count == 1
+            return symbol.id.context.isEmpty
         }
 
         self.relationships = {
--- a/Tests/SwiftDocTests/InterfaceTypeTests.swift
+++ b/Tests/SwiftDocTests/InterfaceTypeTests.swift
@@ -359,4 +359,34 @@ final class InterfaceTypeTests: XCTestCase {
         XCTAssertEqual(members[0].name, "someMethod()")
         XCTAssertEqual(members[1].name, "someExtensionMethod()")
     }
+
+    public func testToplevelSymbols() throws {
+        let source = #"""
+        public class SomeClass {
+            public func someMethod() { }
+        }
+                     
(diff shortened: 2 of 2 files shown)


mattt:
That's totally on me. I introduced this regression by making an unrelated change to `Identifier` path components in 0df8f0a9ae4ddd058a8b43e5a5ce25f7e025c548, and failed to account for current usage.

Thank you for identifying this and adding test coverage. I think better and more extensive testability — especially around what gets built would be valuable. That was one of my goals for #67, but that got away from me in my first few attempts. I'd like to try again after 1.0.0.

As for this, I'm going to apply a hot fix and retag the existing release.

Sources/SwiftDoc/Interface.swift
```diff
@@ -24,7 +24,7 @@ public final class Interface {
                 return false
             }
 
-            return symbol.id.pathComponents.isEmpty
+            return symbol.id.pathComponents.count == 1
```
mattt, line 27:
```suggestion
            return symbol.id.context.isEmpty
```

Tests/SwiftDocTests/InterfaceTypeTests.swift
```diff
@@ -359,4 +359,34 @@ final class InterfaceTypeTests: XCTestCase {
...
+            func someExtensionMethod() { }
+        }
+        """#
+
+        let url = try temporaryFile(contents: source)
+        let sourceFile = try SourceFile(file: url, relativeTo: url.deletingLastPathComponent())
+        let module = Module(name: "Module", sourceFiles: [sourceFile])
+
+        XCTAssertEqual(module.interface.topLevelSymbols.count, 4)
+        print(module.interface.topLevelSymbols)
```
Lukas-Stuehrk, line 385: Oopsie 🤦.

mattt, line 385:
```suggestion
        XCTAssertEqual(module.interface.topLevelSymbols.count, 4)
```

Tests/SwiftDocTests/InterfaceTypeTests.swift
```diff
@@ -359,4 +359,34 @@ final class InterfaceTypeTests: XCTestCase {
...
+
```
mattt, line 386:
```suggestion
        XCTAssertEqual(module.interface.topLevelSymbols.count, 4)

```

mattt (approved):

PR #264, opened 2021-04-26
Fix operator implementation visibility

Lukas-Stuehrk:
The generated documentation could contain operator implementations which should not be included because of their access level.


mattt (approved):

mattt:
Thanks, @Lukas-Stuehrk! Taking a look at #277 now.

PR #219, opened 2021-02-16
Add option for specifying which access levels are included

Lukas-Stuehrk:
This is another try to close #72 - I really want to document my apps.

This slightly changes the behaviour of the internal data structures which collect information on the symbols (e.g. `Interface`, `Members` and `Symbol`). Now, they contain every symbol, not only the public ones. The decision which symbol to include in the documentation is moved to the generating of the documentation.

PR #266, opened 2021-04-28
Add possibility to include private symbols.

Lukas-Stuehrk:
Somehow we discussed this option, documented this option but never implemented this option in #219 🤷.


MaxDesiatov (approved):
Seems legit

PR #271, opened 2021-05-03
build: correct typo

compnerd:
Fix a typo that @MaxDesiatov identified.


MaxDesiatov (approved):
Super frustrating that Xcode 12.5 is not available on GitHub Actions to be able to verify compatibility with this manifest on CI, but this seems legit.

compnerd:
@MaxDesiatov - I don't think that will matter soon.  #269 will be running against 5.5 as that has the prebuilt of `_InternalSwiftSyntaxParser` which is needed here.

PR #272, opened 2021-05-03
git: ignore vim swap files

compnerd:
For those developers who use vim, the traditional behaviour is to create
swap files next to the file itself.  These files should not be committed
to the repository, add them to the ignore list.


compnerd:
Re-apply #255 to get signing by using the web editor.

mattt (approved):

Issue #110, opened 2020-05-12
Expose deep links in the UI


sindresorhus:
For example, in https://sindresorhus.com/playground/KeyboardShortcuts_Name/, it would be nice when I hovered over `init?(raw​Value:​)`, it would show a link icon and clicking it would give me a direct deep link to that method. Like with headers in GitHub readme's.

mattt:
That's a great idea, @sindresorhus.

For posterity, here's a screenshot of how GitHub renders heading anchors:

<img width="206" alt="Screen Shot 2020-05-12 at 12 07 28" src="https://user-images.githubusercontent.com/7659/81734954-2e633b00-9449-11ea-8a3f-2b3163d595ee.png">

mattt:
Fixed by https://github.com/SwiftDocOrg/swift-doc/pull/275

PR #275, opened 2021-05-07
Expose deeplinks.

Lukas-Stuehrk:
This change adds links to the name of all members, so it's easy to copy the direct link. This implements #110.

I chose a different approach than adding a chain symbol when hovering the name. The entire name is a link. This leads to some visual changes in the generated documentation. I can change this behavior to the behavior described in the original issue if wanted.

Before the changes:
Screenshot 2021-05-08 at 00 12 21

After the changes when the entire name becomes the link to the anchor:
Screenshot 2021-05-08 at 00 12 59

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,8 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added support for generating documentation for
   private symbols.
-  #266 by @Lukas-Stuehrk
-  
+  #266 by @Lukas-Stuehrk.
+- Added anchor links to documentation entries for symbols.
+  #275 by @Lukas-Stuehrk.
+
 ### Fixed
 
 - Fixed bug that caused prefix and postfix operators to be omitted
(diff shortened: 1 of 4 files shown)


mattt (approved):

mattt:
This is a nice change, @Lukas-Stuehrk! 

I'd like to give this a quick once-over to get the design a bit closer to what I had in mind with https://github.com/SwiftDocOrg/swift-doc/issues/110#issuecomment-627536084

mattt (approved):
Actually, scratch that — this looks fantastic as-is 🤩 

For whatever reason, the screenshots initially looked off to me. But seeing them in a live render, I wouldn't change a thing. Great work, @Lukas-Stuehrk!

Adding a changelog entry now and merging after that.

Lukas-Stuehrk:
Thanks! 🙇 

Initially, I wanted to build the version as described in #110 and had the current solution as intermediate step. But then I liked the visuals so much that I decided to keep it and go with this solution. I think it looks great and matches the visual language of the index page.

PR #276, opened 2021-05-11
build: bump SwiftSemantics to "main"

compnerd:
This allows the package to resolve with a newer toolchain.  This is
required to get the Windows build enabled.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -74,11 +74,11 @@
         }
       },
       {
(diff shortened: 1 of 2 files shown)


compnerd:
Closing in favour of #278 for signing requirements.  Although that has a better commit message explaining the resolution failure so we have an alternative approach listed there.

Issue #177, opened 2020-09-02
Following links to internal base classes leads to 404 error


rolson:
Doc for internal classes is not generated (good). But the internal ~~base classes and~~ protocol conformances are listed and appear as links. Following these links naturally leads to 404 (bad) as the doc for those types are not generated. 

One proposal is to not make internal types links.

Here is a screen-shot of some ~~inheritances of internal types~~ conformances of internal protocols showing as links:

![image](https://user-images.githubusercontent.com/3718875/92046217-70d5be00-ed3f-11ea-935a-9e14a5c38fe2.png)

mattt:
@rolson My understanding is that public classes can't inherit from non-public superclasses or adopt non-public protocols. Would you mind sharing a minimal example that reproduces this (and also compiles)?

rolson:
You are correct, and I misspoke, sorry. The issue is with internal protocol conformances. 

This compiles and exhibits the issue:

```swift
import Foundation

public class Foo: Bar { }

internal protocol Bar { }
```

Lukas-Stuehrk:
This has been fixed in the latest release ([1.0.0-beta.6](https://github.com/SwiftDocOrg/swift-doc/releases/tag/1.0.0-beta.6)).

PR #277, opened 2021-05-11
Bring back links to other symbols in declarations

Lukas-Stuehrk:
Previous versions of `swift-doc` had a feature which automatically linked other symbols in the declaration of a symbol. This got removed with the changes introduced in #204. I think this is a very useful feature and it seems to be used as the discussion in #177 shows.

This pull requests adds this feature again. It also includes the changes of #264 because it builds upon some changes which were already included there. Because of this, we should merge #264 first before we tackle this pull request.


Sources/SwiftDoc/Interface.swift
```diff
@@ -190,6 +190,14 @@ public final class Interface {
...
         symbolsGroupedByIdentifier.named(name, resolvingTypealiases: resolvingTypealiases)
     }
+
+    public func symbols(named name: String, context: String, resolvingTypealiases: Bool) -> [Symbol] {
+        let foundSymbols = symbols(named: "\(context).\(name)", resolvingTypealiases: resolvingTypealiases)
+        if !foundSymbols.isEmpty {
+            return symbols
+        }
+        return symbols(named: name, resolvingTypealiases: resolvingTypealiases)
+    }
```
mattt, line 200: If the only place we're using this is in `linkTypes`, I think a better option would be to define an extension (`fileprivate` or otherwise) with whatever helper methods you need.

Instead of `symbols(named:context:resolvingTypealiases:)`, you might do:

```swift
extension Collection {
    var nonEmpty: Self? {
        return isEmpty ? nil : self
    }
}

let candidates = (module.interface.symbols(named: "\(context).\(name)", resolvingTypealiases: true).nonEmpty ?? module.interface.symbols(named: name, resolvingTypealiases: true).filter(symbolFilter)
```

Changelog.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   
 ### Fixed
 
+- Fixed bug that caused operator implementations to appear in the documentation
+  although they should be omitted because of their lower access level.
+  #264 by @Lukas-Stuehrk
```
mattt, line 20:
```suggestion
- Fixed links to type declarations.
  #264 by @Lukas-Stuehrk.
- Fixed bug that caused operator implementations to appear in the documentation
  although they should be omitted because of their lower access level.
  #264 by @Lukas-Stuehrk
```

Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -74,9 +74,12 @@ extension SwiftDoc {
           case is Class, is Enumeration, is Structure, is Protocol:
             pages[route(for: symbol)] = TypePage(module: module, symbol: symbol, baseURL: baseURL, includingChildren: symbolFilter)
           case let `typealias` as Typealias:
-            pages[route(for: `typealias`.name)] = TypealiasPage(module: module, symbol: symbol, baseURL: baseURL)
+            pages[route(for: `typealias`.name)] = TypealiasPage(module: module, symbol: symbol, baseURL: baseURL, includingOtherSymbols: symbolFilter)
           case is Operator:
-            pages[route(for: symbol)] = OperatorPage(module: module, symbol: symbol, baseURL: baseURL)
+            let operatorPage = OperatorPage(module: module, symbol: symbol, baseURL: baseURL, includingImplementations: symbolFilter)
+            if !operatorPage.implementations.isEmpty {
```
mattt, line 80: Is the thinking here that operators without implementations should be considered private/reserved and therefore documented? Or that we don't handle this case well?

Sources/SwiftDoc/Interface.swift
Lukas-Stuehrk: Agree. There's some chance that we might need something like this also at other places in the future. But then we should generalize when we have the problem and not earlier.

I like your suggestion and will apply it.

mattt (approved):

PR #282, opened 2021-05-18
Style callout delimiters in discussion for HTML output

### LINKED ISSUE #281: Add designs to accommodate all types of callout in discussion in HTML output
[Swift Markup](https://github.com/SwiftDocOrg/SwiftMarkup) supports a variety of callouts, including `Warning`, `Note`, `Precondition` and `SeeAlso`. At the moment, we aren't doing much to visually distinguish these in HTML output. It'd be great to give these some panache ✨ 

Screen Shot 2021-05-18 at 09 51 56

mattt:
Fixes #281 

This PR adopts the following color coding for annotation types, but I'm very much open to changing this around:

- **Error** (`bug`): Pink
- **Warning** (`attention`, `important`, `warning`): Orange
- **Metadata** (`author[s]`, `copyright`, `date`): Gray
- **Informational** (_everything else_): Blue

This doesn't handle custom callouts, but those are Xcode Playgrounds-specific, and the default rendering is good enough.

* * *

Screen Shot 2021-05-18 at 10 20 51
Screen Shot 2021-05-18 at 10 20 33
Screen Shot 2021-05-18 at 10 21 00
Screen Shot 2021-05-18 at 10 20 25


Lukas-Stuehrk (commented):
Nice addition, @mattt. This is one of the features where you immediately realise how much this was missing before! This makes the use of callouts so much clearer and useful – and also more familiar for users of Apple's documentations.


I think we should change the color values a little bit. I find the general color scheme and which color is used good and reasonable. But for all three callout styles, the rendered text has a contrast ratio to its background of 1 or slightly above. This is not very nice for people with visual impairment. We should try to go with a contrast ratio of at least 4.5.

mattt:
@Lukas-Stuehrk Thanks for pointing out the contrast ratios for that and suggesting more accessible alternatives. These look a lot more reasonable:

Screen Shot 2021-06-01 at 11 36 09

mattt:
Still some room for improvement, possibly using text shadows to preserve contrast without compromising the color palette as much. And some work to do to make this work with dark mode (#160 I know, sorry for the delay 🙇). But I think this is good enough for now. :shipit:

PR #287, opened 2021-05-22
Add new parameter for date formats, and default to en_US

f-meloni:
Hello everyone 👋🏻 

First of all, thank you for this great tool!

While I was generating some documentation I've noticed that it was using my locale, which makes the result odd, because uses an italian word in the middle of an english sentence.

Schermata 2021-05-22 alle 20 58 53

I've then added a new option (that defaults to `en_US`) to specify the locale for the dates that are printed by swift docs.
I could just hardcode `en_US` there, but I thought that making it explicit with an option and allow the user to modify it was a better choice.

Schermata 2021-05-22 alle 21 32 42

And now is correct also with my italian locale

Schermata 2021-05-22 alle 21 47 38

### DIFF
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -51,7 +51,7 @@ extension SwiftDoc {
     var options: Options
 
     func run() throws {
-     let datesLocale = Locale(identifier: self.options.datesLocale)
+     let datesLocale = Locale(identifier: options.datesLocale)
     
       for directory in options.inputs {
         var isDirectory: ObjCBool = false
--- a/Sources/swift-doc/Subcommands/Generate.swift
+++ b/Sources/swift-doc/Subcommands/Generate.swift
@@ -132,11 +132,29 @@ extension SwiftDoc {
         } else {
           switch format {
           case .commonmark:
-            pages["Home"] = HomePage(module: module, externalTypes: Array(symbolsByExternalType.keys), baseURL: baseURL, datesLocale: datesLocale, symbolFilter: symbolFilter)
-            pages["_Sidebar"] = SidebarPage(module: module, externalTypes: Set(symbolsByExternalType.keys), baseURL: baseURL, datesLocale: datesLocale, symbolFilter: symbolFilter)
+            pages["Home"] = HomePage(
+                module: module,
+                externalTypes: Array(symbolsByExternalType.keys),
+                baseURL: baseURL,
+                datesLocale: datesLocale,
+                symbolFilter: symbolFilter
+            )
+            pages["_Sidebar"] = SidebarPage(
+                module: module,
+                externalTypes: Set(symbolsByExternalType.keys),
+                baseURL: baseURL,
+                datesLocale: datesLocale,
+                symbolFilter: symbolFilter
+            )
             pages["_Footer"] = FooterPage(baseURL: baseURL, datesLocale: datesLocale)
           case .html:
-            pages["Home"] = HomePage(module: module, externalTypes: Array(symbolsByExternalType.keys), baseURL: baseURL, datesLocale: datesLocale, symbolFilter: symbolFilter)
+            pages["Home"] = HomePage(
+                module: module,
+                externalTypes: Array(symbolsByExternalType.keys),
+                baseURL: baseURL,
+                datesLocale: datesLocale,
+                symbolFilter: symbolFilter
+            )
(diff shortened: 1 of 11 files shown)


Sources/swift-doc/Supporting Types/Pages/SidebarPage.swift
```diff
@@ -2,6 +2,7 @@ import SwiftSemantics
 import SwiftDoc
 import CommonMarkBuilder
 import HypertextLiteral
+import struct Foundation.Locale
```
f-meloni, line 5: In some files the compiler was having some issues with `Protocol` once Foundation was imported, also when I used `SwiftSemantics.Protocol` instead of just `Protocol`, so I've imported only `Foundation.Locale` where the issue was happening

Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -47,6 +51,8 @@ extension SwiftDoc {
     var options: Options
 
     func run() throws {
+     let datesLocale = Locale(identifier: self.options.datesLocale)
```
MaxDesiatov, line 54: I don't think the codebase uses `self.` where it isn't needed, so I suggest omitting it here for consistency.

```suggestion
     let datesLocale = Locale(identifier: options.datesLocale)
```

Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -72,11 +78,11 @@ extension SwiftDoc {
         for symbol in module.interface.topLevelSymbols.filter(symbolFilter) {
           switch symbol.api {
           case is Class, is Enumeration, is Structure, is Protocol:
-            pages[route(for: symbol)] = TypePage(module: module, symbol: symbol, baseURL: baseURL, includingChildren: symbolFilter)
+            pages[route(for: symbol)] = TypePage(module: module, symbol: symbol, baseURL: baseURL, datesLocale: datesLocale, includingChildren: symbolFilter)
```
MaxDesiatov, line 81: Please avoid long lines, as with these changes it now requires horizontal scrolling to review even on a wide screen, and makes it even harder on narrow screens. This is applicable to the rest of the long lines in this diff.

```suggestion
            pages[route(for: symbol)] = TypePage(
              module: module,
              symbol: symbol,
              baseURL: baseURL,
              datesLocale: datesLocale,
              includingChildren: symbolFilter
            )
```

Sources/swift-doc/Subcommands/Generate.swift
f-meloni: Good call, thank you!
I started with `let datesLocale = Locale(identifier: self.datesLocale)`, because I forgot it was inside options, and then forgot to remove the self 😅

Sources/swift-doc/Subcommands/Generate.swift
f-meloni: Would be cool to some swiftformat and swiftlint config to automate this :P
What do you think?
I can help to set it up if you want

MaxDesiatov: Thank you, sounds good to me! @mattt WDYT?

Sources/swift-doc/Supporting Types/Pages/FooterPage.swift
```diff
@@ -37,6 +40,8 @@ struct FooterPage: Page {
     }
 
     var html: HypertextLiteral.HTML {
+        dateFormatter.locale = datesLocale
```
f-meloni, line 43: I've also noticed that there is a discrepancy with the Markdown footer, that uses `"Generated at \(timestamp) using [swift-doc](\(href)) \(SwiftDoc.configuration.version)."` while it could use `"Generated on \(dateString)"` instead.
WDYT?

Sources/swift-doc/Subcommands/Generate.swift
compnerd: If we are going to add linting, I would suggest that we should stick to swift-format for linting rather than using swift-lint.  As to setting that up, its pretty trivial (there is an example at https://github.com/compnerd/swift-win32/blob/main/.github/workflows/lint.yml).

mattt (commented):
Thanks so much for taking the time to put this PR together, @f-meloni. I'd be interested to explore a full localization strategy for `swift-doc` (#290). For now, I think the best solution would be to configure date formatters to use an `en_US` locale (#289).

PR #288, opened 2021-05-26
Embed generated CSS in Swift raw string literal

### LINKED ISSUE #231: CSS assets don't reflect swift-doc's versioning
Currently, swift-doc generates the CSS assets by [simply downloading the latest version of the minified CSS file from GitHub](https://github.com/SwiftDocOrg/swift-doc/blob/467cf38240f53f48129a9ef3e7ac969b939a7d15/Sources/swift-doc/Subcommands/Generate.swift#L141-L141). This will break old versions once we keep iterating on the CSS and introduce backwards-incompatible changes. 

Therefore we should find a way to use the correct version of the CSS files that relates to the version of swift-doc which is used to generate the documentation.

I think the easiest solution is to switch back to get the CSS from the package resources. There was already an attempt in #192. Unfortunately this introduced a lot of issues and needed to be reverted. In the meantime, the minimal Swift version for swift-doc was bumped to Swift 5.3, so from my understanding this approach should be safe now.

mattt:
Resolves #231 
Alternative to #267 

As discussed in https://github.com/SwiftDocOrg/swift-doc/pull/267#issuecomment-831939559, this PR takes an alternative approach. Rather than trying to make Swift package resources work, this PR changes our asset pipeline to write directly to a raw string literal constant in a Swift source file, which can be referred to directly in code.


Sources/swift-doc/Subcommands/Generate.swift
```diff
@@ -150,9 +150,9 @@ extension SwiftDoc {
         }
 
         if case .html = format {
-          let cssData = try fetchRemoteCSS()
-          let cssURL = outputDirectoryURL.appendingPathComponent("all.css")
-          try writeFile(cssData, to: cssURL)
+            let cssData = css.data(using: .utf8)!
+            let cssURL = outputDirectoryURL.appendingPathComponent("all.css")
+            try writeFile(cssData, to: cssURL)
```
MaxDesiatov, line 155: nit: 4 spaces indentation looks inconsistent here with the rest of the file

MaxDesiatov (approved):
Overall LGTM, just a minor nitpick

.node/package.json
```diff
@@ -5,15 +5,16 @@
     "not dead"
   ],
   "scripts": {
-    "watch": "postcss -w ../Assets/**/*.css -o ../Resources/all.min.css --config ./postcss.config.js",
-    "build": "postcss ../Assets/**/*.css -o ../Resources/all.min.css --config ./postcss.config.js"
+    "build": "node index.js",
+    "watch:build": "nodemon -w ../Assets/ -e css -x \"npm run build\"",
+    "watch": "npm run watch:build"
```
Lukas-Stuehrk, line 10: 🤔 Could you elaborate what's the idea of this indirection?

Lukas-Stuehrk (approved):

Sources/swift-doc/Subcommands/Generate.swift
mattt: Thanks for catching that. All the more reason to integrate a linter / formatter as discussed in https://github.com/SwiftDocOrg/swift-doc/pull/287#discussion_r637529830.

.node/package.json
mattt: You're right, this extra level of indirection isn't doing anything. Fixed by 9961480.

PR #295, opened 2021-06-02
Update publish.yml

compnerd:
Workaround the product versioning style for packaging on Windows.


compnerd:
CC: @mattt - what do you think of this as a workaround for the versioning issue?  Could you please cut another release?

